PR #10, opened 2020-10-19
New pallet to mint USDv

c410-f3r:
* Users that are part of `pallet_membership::DefaultInstance` are providers.

* Changes `CurrencyId` type from `u32` to `ValiuCurrencies` for better maintainability and expressiveness.

* A new crate called `valiu-node-commons` was created to store common configurations across pallets, runtime and node.

* Users that are part of `pallet_membership::Instance0` can mint USDv.

* Runs Rustfmt for consistency and to make life easier when writing new code


pallets/liquidity-provider/src/lib.rs
```diff
@@ -47,7 +45,11 @@ decl_module! {
         fn deposit_event() = default;
 
         #[weight = 0]
-        pub fn attest(origin, asset_id: CurrencyIdOf<T>, balance: BalanceOf<T>)   -> dispatch::DispatchResult{
+        pub fn attest(origin, asset_id: ValiuCurrencies, balance: BalanceOf<T>) -> dispatch::DispatchResult {
+            if let ValiuCurrencies::Usdv = asset_id {
+                return Err(Error::<T>::MustNotBeUsdv.into());
+            }
```
c410-f3r, line 51: This is a bit of a hack. I still need to find a better way to handle this properly

olanod: I think it's fine for now :) Later(next ticket?) we could group currencies in different ways based on the usage and there wouldn't be a need for a check if the type is explicit about the allowed currency. I'm not sure if the SCALE encoder supports it but our currencies enum can be an union of other enums, so when we do need to work with all the currencies we have something that groups them but when we don't we use the specific variant we need.
```rust
enum Asset {
    USDv,
    OffChain(OffChainAsset), // for attestations
    Collateral(Collateral) // assets valid for use as USDv collateral?
    // or Fiat(Fiat), Digita(Digital)
    // ... or other ways of grouping? 🤷🏽‍♂️ 
}
```

valiu-node-commons/src/valiu_currencies.rs
```diff
@@ -0,0 +1,35 @@
...
+    Copy,
+    Debug,
+    Eq,
+    Ord,
+    PartialEq,
+    PartialOrd,
+    parity_scale_codec::Decode,
+    parity_scale_codec::Encode,
+)]
+pub enum ValiuCurrencies {
```
olanod, line 15: Complementing the other comment, what do you say about in general using better the word asset instead of currency since it's a bit more general and we don't know if our currency will be backed by Dollars, houses or potatoes?

olanod (commented):
Nice! looks good to me. What do you think about the grouping of assets?

valiu-node-commons/src/valiu_currencies.rs
c410-f3r: Hehehehe, potatoes was funny.

Sure. `Asseet` will be

pallets/liquidity-provider/src/lib.rs
c410-f3r: Nice idea!

PR #11, opened 2020-10-19
Tweak CI script

c410-f3r:
Taking the last PR into consideration, `Rustfmt` is now enabled on CI.

Also updates `actions/cache` to version 2.x and runs quick checks on pull requests.


olanod (commented):
Very much needed :) `cargo test` as well?

PR #12, opened 2020-10-20
Lock attested assets

c410-f3r:
Every new amount of attested assets ares now automatically moved into the "reserved" set of balances


olanod (approved):
Simple enough :)

PR #15, opened 2020-10-21
Mint USDv based on assets of providers

c410-f3r:
It is not clear yet how mint should be handled, nevertheless, this is an attempt on T-198 to give an overall architectural idea


pallets/usdv-minting/src/lib.rs
```diff
@@ -42,14 +47,30 @@ decl_module! {
...
-            let members = pallet_membership::Module::<T, Instance0>::members();
-            members.binary_search(&provider).ok().ok_or(pallet_membership::Error::<T, Instance0>::NotMember)?;
+            if balance > T::Currency::reserved_balance(asset_id, &provider_member) {
+                return Err(Error::<T>::InsufficientFunds.into());
+            }
 
-            T::Currency::deposit(Asset::Usdv, &provider, balance)?;
-            Self::deposit_event(RawEvent::Mint(provider, Asset::Usdv));
+            T::Currency::deposit(Asset::Usdv, &provider_member, balance)?;
+            Self::deposit_event(RawEvent::Mint(mint_member, Asset::Usdv));
```
c410-f3r, line 73: Basically searches if a given provider has funds that can be converted into USDv. Since "attested" assets are already locked by default, this function simply checks for funds and adds Valiu Dolars accordingly

pallets/usdv-minting/src/lib.rs
```diff
@@ -42,14 +47,34 @@ decl_module! {
...
-            let members = pallet_membership::Module::<T, Instance0>::members();
-            members.binary_search(&provider).ok().ok_or(pallet_membership::Error::<T, Instance0>::NotMember)?;
+            if balance > diff_amount {
+                return Err(Error::<T>::InsufficientFunds.into());
+            }
 
-            T::Currency::deposit(Asset::Usdv, &provider, balance)?;
-            Self::deposit_event(RawEvent::Mint(provider, Asset::Usdv));
+            T::Currency::deposit(Asset::Usdv, &provider_member, balance)?;
+            Self::deposit_event(RawEvent::Mint(mint_member, Asset::Usdv));
```
c410-f3r, line 77: Basically searches if a given provider has funds that can be converted into USDv. Since "attested" assets are already locked by default, this function simply checks for funds and adds Valiu Dolars accordingly

pallets/usdv-minting/Cargo.toml
```diff
@@ -9,6 +9,7 @@ valiu-node-commons = { default-features = false, path = '../../valiu-node-common
 
 [dev-dependencies]
 orml-tokens = { default-features = false, version = '0.3' }
+pallet-provider = { default-features = false, path = '../liquidity-provider' }
```
olanod, line 12: This got me thinking about what you mentioned at some point of having a single pallet instead of splitting into two. Having to use the provider pallet like this is a design smell, probably the two things belong together. With the way we are going to follow it probably makes sense to make the minting part of what a provider does, to keep both decoupled it would require that attested assets go to a central pot(some configurable accountId) where the "minters" would take the collateral to create USDv.  
Anyway sorry that the design has been a bit uncertain but now I have more clear that USDv creation should be part of the functions of the liquidity provider and actually I think it should be done automatically as soon as a provider makes an attestation with a "collateral asset".

Imagining it can be something like [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=0646cc3f7e4edfddd4ebf88f2620ce1b)

pallets/usdv-minting/src/lib.rs
```diff
@@ -13,14 +12,20 @@ mod tests;
 
 type BalanceOf<T> =
     <<T as Trait>::Currency as MultiCurrency<<T as frame_system::Trait>::AccountId>>::Balance;
+type MintMembers = pallet_membership::Instance0;
```
olanod, line 15: So if moving minting to the provider pallet we would still have this membership as a configurable associated type that is initially also set to the default instance so every provider can be minter initially but wouldn't always be the case as all minters are providers but not all providers are minters.

olanod (changes requested):
Oii! Nice job there! Sorry it took me a while and that the requirements haven't been that clear. Check the comment about merging the pallets and ping me for further discussion or if there's something you don't agree with

pallets/liquidity-provider/src/lib.rs
```diff
@@ -3,18 +3,29 @@
...
+    pallet_membership::Trait<MintMembers> + pallet_membership::Trait<ProviderMembers>
+{
+    type Asset: MultiCurrency<
+        Self::AccountId,
+        Balance = <<Self as Trait>::Collateral as MultiCurrency<
+            <Self as frame_system::Trait>::AccountId,
+        >>::Balance,
+        CurrencyId = Asset,
+    >;
+    type Collateral: MultiReservableCurrency<Self::AccountId, CurrencyId = Asset>;
```
c410-f3r, line 28: The `token` module expects a single `CurrencyId` but in the near future, `Collateral` should be `MultiReservableCurrency<Self::AccountId, CurrencyId = Collateral>` somehow.

olanod (approved):

PR #16, opened 2020-10-27
Transfer collaterals

c410-f3r:


valiu-node-commons/src/distribution_strategy.rs
```diff
@@ -0,0 +1,33 @@
...
+    Eq,
+    Ord,
+    PartialEq,
+    PartialOrd,
+    parity_scale_codec::Decode,
+    parity_scale_codec::Encode,
+)]
+pub enum DistributionStrategy {
+    Evenly,
+}
```
c410-f3r, line 17: It is possible to split collaterals by many ways and this is just one of them

olanod: Yeah that's nice! :D that's how I thought it could be done 👍🏽

pallets/liquidity-provider/src/transfer_handlers.rs
```diff
@@ -0,0 +1,164 @@
...
+    }
+    T::Asset::transfer(Asset::Usdv, &from, &to, amount)?;
+    Module::<T>::deposit_event(RawEvent::Transfer(from, to, amount));
+    Ok(())
+}
+
+// Just check if a given set of collateral balances can fulfil the desired amount of USDv by
+// subtracting each asset
+#[inline]
+fn test_transfer_evenly<T>(
```
olanod, line 112: Since we mint USDv 1:1 with the collateral, would it be safe to assume USDv balance is in sync with its collateral, so it would be enough to check at the beginning if I can transfer the USDv then I can transfer the collateral?

olanod (commented):
Nice job! not a trivial thing to implement. Let me know what you think

pallets/liquidity-provider/src/transfer_handlers.rs
c410-f3r: The additional checking is a plus since this function is still necessary to correctly calculate how much each collateral will be subtracted.

PR #17, opened 2020-10-28
Maintenance bits

c410-f3r:
Each commit should be read individually

### DIFF
--- a/pallets/liquidity-provider/src/lib.rs
+++ b/pallets/liquidity-provider/src/lib.rs
@@ -33,7 +33,8 @@ decl_event!(
         AccountId = <T as frame_system::Trait>::AccountId,
         Balance = Balance<T>,
     {
-        Attestation(AccountId, Collateral),
+        Attestation(AccountId, Asset),
+        Mint(AccountId, Collateral),
         Transfer(AccountId, AccountId, Balance),
     }
 );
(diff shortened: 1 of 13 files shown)


pallets/liquidity-provider/src/lib.rs
```diff
@@ -56,26 +57,30 @@ decl_module! {
         #[weight = 0]
         pub fn attest(
```
olanod, line 58: What happens when we attest `USDv`, seems possible but it shouldn't.

pallets/liquidity-provider/src/lib.rs
```diff
@@ -56,26 +57,30 @@ decl_module! {
...
+            asset: Asset,
             balance: Balance<T>
         ) -> dispatch::DispatchResult
         {
             let who = ensure_signed(origin)?;
+            do_attest::<T>(who, asset, balance)
+        }
 
+        #[weight = 0]
+        pub fn mint(
```
olanod, line 69: Still not sure about exposing this as a public dispatchable call, maybe as a private function of the module?

c410-f3r: mint method was removed alongside with mint members

pallets/liquidity-provider/src/lib.rs
```diff
@@ -56,27 +55,21 @@ decl_module! {
...
+            match asset {
+                Asset::Usdv => return Err(crate::Error::<T>::MustNotBeUsdv.into()),
+                Asset::Collateral(collateral) => {
+                    let who = ensure_signed(origin)?;
+                    do_attest::<T>(who.clone(), Asset::Usdv, balance)?;
+                    T::Collateral::deposit(collateral.into(), &who, balance)?;
+                    T::Collateral::reserve(collateral.into(), &who, balance)?;
+                    Self::deposit_event(RawEvent::Attestation(who, collateral.into()));
+                    Ok(())
+                }
```
c410-f3r, line 71: Since every attestation is also a mint, everything that is not a stable collateral will fail

PR #18, opened 2020-10-29
Provide and update a list of offers

c410-f3r:
Depends on #17


olanod (approved):
LGTM. One final thought to keep in mind though, storing rates as `[provider][asset] => {asset, rate}` I guess is easier and cheaper for updating the rates but if later we start doing a lot of querying by asset pair it might be expensive to look up wouldn't it? maybe later we also store rates as `[asset][asset] => {provider, rate}` to query easy by asset?

c410-f3r:
@olanod Changed to [asset][asset]

olanod (approved):

PR #22, opened 2020-11-14
Weights for method calls

c410-f3r:
This PR uses the default Substrate benchmark framework to calculate a upper-bound weight for each public method call.

Weight is a unit of time and every signed extrinsic should have a weight to calculate fees or the average number of transactions for a block. The same necessity is true for unsigned transactions because theoretically anyone could submit infinite number of extrinsics inside a single block for anything that has `#[weight = 0]`.

Please, keep in mind that the current generated weights are based on my machine (Notebook with a i5-7200U processor) but the actual values should be generated from a reference hardware.

```bash
cargo run --release --features runtime-benchmarks -- benchmark --chain testnet_testnet --steps 50 --repeat 20 --pallet pallet_liquidity_provider --extrinsic "*" --wasm-execution=compiled --execution=wasm --output
```

To accomplish the benchmark structure, imports were re-arranged and more tests were included to enhance robustness. 

#### References:

https://substrate.dev/docs/en/knowledgebase/learn-substrate/weight
https://github.com/paritytech/substrate/tree/master/frame/benchmarking

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -44,6 +44,4 @@ jobs:
           command: clippy
           args: --all-features -- -D warnings
 
-      - uses: actions-rs/cargo@v1
-        with:
-          command: test
+      - run: ./scripts/checks.sh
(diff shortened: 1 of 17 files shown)


olanod:
Looks good to me, I'll merge now and once mainnet validators are running I'll make sure the weights are updated so it's consistent with the production hardware.

PR #26, opened 2020-12-21
Remove usage of pallet_balances

c410-f3r:


valiu-node-runtime-types/src/lib.rs
```diff
@@ -81,7 +84,7 @@ pub type SignedExtra<R> = (
     CheckEra<R>,
     CheckNonce<R>,
     CheckWeight<R>,
-    ChargeTransactionPayment<R>,
+    DummySignedTransaction<R>,
```
olanod, line 87: Seems like this might not be needed at all, `SignedExtra` trait is implemented for tuples(1-12 elements) where each element implements the trait itself, adding more elements is just adding more checks for the transactions, adding a dummy check probably is the same as having no check at all.

olanod (changes requested):

Issue #30, opened 2020-12-24
Add "Transfer" remote call


olanod:
"Transfer" is the main call we'll be using to transfer USDv and later any other token. Needs to be exposed via RPC.

PR #32, opened 2020-12-24
Add attest and transfer external calls

### LINKED ISSUE #31: Add "Attest" remote call
`Attest` is how USDv is initially minted, a provider "attesting" collateral X automatically creates the same amount of USDv. This needs to be exposed via RPC.

c410-f3r:
Fixes #31 and #30


c410-f3r:
The removal of `ChargeTransactionPayment` from `Extra` broke the default extrinsic serialization/deserialization and it was causing a lot of weird and unknown errors. To workaround this limitation, a manual `Extra` implementation was added for RPC interaction.

valiu-node-rpc/src/liquidity_provider.rs
```diff
@@ -0,0 +1,36 @@
...
+    pub from: <T as System>::AccountId,
+    pub to: <T as System>::AccountId,
+    pub balance: Balance,
+}
+
+#[derive(Clone, Debug, PartialEq, parity_scale_codec::Encode, substrate_subxt::Call)]
+pub struct TransferCall<T: LiquidityProvider> {
+    pub to: <T as System>::AccountId,
+    pub to_amount: Balance,
+    pub ds: DistributionStrategy,
```
olanod, line 35: Let's remove this guy and if it's not set as default already let's make it the default, I prefer to keep the client facing API simpler and besides the distribution strategy is something that would eventually be changed for everyone via governance or based on some preferences of the recipient that for whatever reason doesn't want or can not receive USDv backed by a specific asset.

valiu-node-rpc/src/valiu_extra.rs
```diff
@@ -0,0 +1,179 @@
+use core::{fmt, marker::PhantomData};
```
olanod, line 1: I'm fine with this custom "`Extra`" implementation but would you say the weird unknown errors are bugs in any of the dependencies or something we should add a note and possibly create issues somewhere to track for future reference and hopefully remove this custom implementation some day?

olanod (changes requested):
Thanks for this! very much needed :D just a small change and a some comments

runtime/src/lib.rs
```diff
@@ -398,16 +389,16 @@ impl orml_tokens::Trait for Runtime {
 #[derive(Debug)]
 pub struct OffchainAppCrypto;
 
-impl frame_system::offchain::AppCrypto<MultiSigner, MultiSignature> for OffchainAppCrypto {
-    type GenericPublic = sp_runtime::app_crypto::sr25519::Public;
-    type GenericSignature = sp_runtime::app_crypto::sr25519::Signature;
+impl frame_system::offchain::AppCrypto<AccountId, Signature> for OffchainAppCrypto {
```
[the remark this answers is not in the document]
c410-f3r: The `MultiSignature` enum came from the official template and I removed it because: (1) The pallet runtime (`sr25519`) was working fine while the RPC runtime (`MultiSignature`) was not. (2) It was easier to satisfy all trait constraints of all different runtimes using `sr25519` and (3) We currently don't require different types of signature (`sr25519`, `ed25519` or `ecdsa`).

If desirable, I can try to reintroduce `MultiSignature` in a following PR

valiu-node-rpc/src/liquidity_provider.rs
```diff
@@ -0,0 +1,36 @@
...
+
+#[substrate_subxt::module]
+pub trait LiquidityProvider: System {}
+
+#[derive(Clone, Debug, PartialEq, parity_scale_codec::Encode, substrate_subxt::Call)]
+pub struct AttestCall<T: LiquidityProvider> {
+    pub asset: Asset,
+    pub balance: Balance,
+    pub offer_rates: Vec<OfferRate<Balance>>,
+    pub phantom: PhantomData<T>,
```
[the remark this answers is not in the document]
c410-f3r: Yeah, `substrate_subxt::Call` procedural macro expects a struct with a `T` type, otherwise, it won't work

valiu-node-rpc/src/valiu_extra.rs
c410-f3r: Indeed, this feature should be noted somewhere since Substrate rejects extrinsics sent using the default extra parameters.
I personally don't know how to remove this custom `Extra` without defaulting back to the original type. Maybe there is a Substrate parameter that deals with it but I am not sure.

scripts/valiu-node-rpc-integration-tests.sh
```diff
@@ -1,10 +1,11 @@
 #!/usr/bin/env bash
 
-set -euxo pipefail
+set -uxo pipefail
```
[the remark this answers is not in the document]
c410-f3r: Cool!

runtime/src/lib.rs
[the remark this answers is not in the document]
olanod: I don't have enough knowledge to know if we would like to reintroduce it and what difference would it make for the offchain worker keys. If you see a compelling usage we can create a ticket to be able to track it an later re-introduce it otherwise let's leave it like that.

valiu-node-rpc/src/valiu_extra.rs
olanod: Then if is not to much to ask could file an issue in substrate or the relevant repo and add a `// NOTE: <link>` here for future reference? in a few months it would be forgotten this is a "hack" and just remain there forever as part of the landscape 😜

c410-f3r: Hum... I am sorry for the misunderstanding, Substrate is not the one to blame here because it is only trying to de-serialize what was defined in the runtime. The truth is: `DefaultExtra` of `substrate-subxt` contains `TransactionPayment` but our runtime doesn't expect it because `pallet-balance` was  removed, thus the need to define our own `ValiuExtra` to successfully send extrinsics. It is more like a feature then a "hack".

To remove some burden on our side, I will try to re-export all these duplicated `Extra` types ìn `substrate-subxt`.

olanod: Ok now I get it! 👍🏽 and sure feel free to make it look prettier :)

c410-f3r:
@olanod Can you merge this PR? The `substrate-subxt` changes will take some time until the Parity guys deploy a new version.

c410-f3r:
Thanks!

PR #35, opened 2020-12-25
Add workflow badges

stanly-johnson:
Add workflow status badges to README [Preview](https://github.com/stanly-johnson/vln-node/blob/build-badge/README.md)


olanod:
We'll check why the test fails but thanks for the addition!

PR #36, opened 2020-12-25
remove properties from chainspec

stanly-johnson:
fix: since balances pallet has been removed(#26), setting properties in chainSpec does not serve a purpose anymore

### DIFF
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -2,7 +2,6 @@ pub(crate) mod dev;
 pub(crate) mod local_testnet;
 pub(crate) mod testnet_testnet;
 
-use serde_json::Map;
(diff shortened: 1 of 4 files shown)


node/src/chain_spec.rs
```diff
@@ -2,7 +2,6 @@ pub(crate) mod dev;
 pub(crate) mod local_testnet;
 pub(crate) mod testnet_testnet;
 
-use sc_service::Properties;
 use serde_json::Map;
```
olanod, line 5: Check the unused import :)

olanod (changes requested):

node/src/chain_spec.rs
stanly-johnson: fixed

PR #47, opened 2021-01-10
Remove `valiu-node-runtime-types` and add some docs

c410-f3r:
In every place that `valiu-node-runtime-types` was being used, `valiu-node-commons` was also being used so both are now merged.

Also adds some documentation


olanod (changes requested):
Nice! Love "house cleaning" and simplifying the project structure ❤️ Now that we are on it, one small change to reflect direction of the project... The idea is to replace mentions of Valiu for just "VLN", I'd like to keep the distinction between the _Valiu Liquidity Network_(the name of the blockchain) and ignore Valiu's existence since it will be just another user of the system(a liquidity provider). So the module would be something like `vln_commons`

c410-f3r:
@olanod Everything related to Valiu has been renamed to `vln` or `Valiu liquidity network`

PR #54, opened 2021-02-05
fix: handle offchain destinations

### LINKED ISSUE #43: Destination can be off-chain
A transfer destination will not always be an account within the blockchain, it can for example be a bank account or an account in a different blockchain that a liquidity provider controls and has notified the VLN of its existence.

This means `to` parameter of the `Transfer` call should change from an AccountId to an enum type that allows other types of off-chain destinations.
e.g.
```rust
enum Destination {
    Vln(AccountId),
    Bank(DetailsId),
    Btc(BtcId),
    // ...
}
```

stanly-johnson:
Resolves #43

Support for possible offchain destinations. Perform onchain transfer if the destination is a supported address, else handle according to destination type. 

### Changes
- [x] Add destination enum
- [x] Modify transfer extrinsic
- [x] Tests and documentation


stanly-johnson:
@olanod can you take a look and let me know if this is the right way to go. If the destination is offchain I am just rejecting it. I guess we can emit events, that would be a better look? Also any more destinations to support?

commons/src/destination.rs
```diff
@@ -0,0 +1,54 @@
...
+    PartialEq,
+    PartialOrd,
+    parity_scale_codec::Decode,
+    parity_scale_codec::Encode,
+)]
+pub enum Destination {
+    // type to denote destinations on vln blockchain
+    Vln(AccountId),
+    // type to denote transactions on the bitcoin blockchain
+    Btc(BtcTransactionId),
```
olanod, line 24: This was mostly to give an example so I'd leave it out for now since we are not likely to support BTC destinations in the time being.

commons/src/destination.rs
```diff
@@ -0,0 +1,54 @@
...
+    // type to denote off chain bank transfer transations
+    Bank(BankTransactionId),
```
olanod, line 26: This on the other hand will be key to support it soon, since bank information will contain personal user information that shouldn't touch the blockchain this will likely be an IPFS CID(an u64?) that refers to an encrypted document on IPFS that only the final liquidity provider on the chain can decrypt to send funds to the user's account

pallets/liquidity/src/lib.rs
```diff
@@ -135,16 +136,23 @@ decl_module! {
...
+                // preform onchain transfer for vln address
+                Destination::Vln(to_address) => {
+                    let from = ensure_signed(origin)?;
+                    if to_amount.is_zero() {
+                        return Err(crate::Error::<T>::TransferMustBeGreaterThanZero.into());
+                    }
+                    Self::transfer_evenly(from, to_address.into(), to_amount)?;
+                    Ok(())
+                }
+                _ => Err(crate::Error::<T>::DestinationNotSupported.into())
```
olanod, line 154: Yeah this should work for now 👍🏽

commons/src/destination.rs
```diff
@@ -0,0 +1,50 @@
+use core::fmt;
+use sp_core::H256;
+
+// BankTransactionCid should hold a sha256 cid from ipfs
+pub type BankTransactionCid = H256;
```
olanod, line 5:
```suggestion
pub type BankDetailsCid = H256;
```

commons/src/destination.rs
```diff
@@ -0,0 +1,50 @@
+use core::fmt;
+use sp_core::H256;
+
+// BankTransactionCid should hold a sha256 cid from ipfs
```
olanod, line 4:
```suggestion
// BankDetailsCid should hold a sha256 cid from ipfs
```

commons/src/destination.rs
```diff
@@ -0,0 +1,50 @@
...
+    PartialEq,
+    PartialOrd,
+    parity_scale_codec::Decode,
+    parity_scale_codec::Encode,
+)]
+pub enum Destination<AccountId> {
+    // type to denote destinations on vln blockchain
+    Vln(AccountId),
+    // type to denote off chain bank transfer transations
+    Bank(BankTransactionCid),
```
olanod, line 24:
```suggestion
    Bank(BankDetailsCid),
```

PR #56, opened 2021-02-09
feat: Add proxy pallet

### LINKED ISSUE #51: Add proxy pallet
Adding support for proxy accounts will simplify the way Valiu as a liquidity provider manages the accounts of its users.
This is important for organizations joining as liquidity providers since they have more complex needs on how funds should be handled.

stanly-johnson:
Resolves #51 

Adds [proxy-pallet](https://github.com/paritytech/substrate/tree/v2.0.0/frame/proxy) to vln node


stanly-johnson:
Hi @olanod, the proxy-pallet is designed for a [base reservable currency](https://github.com/paritytech/substrate/blob/a200cdb93c6af5763b9c7bf313fa708764ac88ca/frame/proxy/src/lib.rs#L84), the token pallet implements it [differently](https://github.com/open-web3-stack/open-runtime-module-library/blob/8c150f34a5b305103438e93337e1b9e7868dd1f3/tokens/src/lib.rs#L988) (using token type). 

We can either add the implementation, fixing the token-id to `USDv` or copy the `proxy-pallet` code and make it support token pallet. What do you suggest?

Also, current implementation supports any `Call` to be proxied, should this be restricted? Maybe just to transfer operations?

olanod:
I don't fully understand the limitation of the implementation of the tokens pallet but copying the `proxy-pallet` code seems quite ugly. I'd like to see more how the "fixing the token-id" would be :)

stanly-johnson:
My bad, the integration is pretty straightforward, no hacks needed :)

runtime/src/lib.rs
```diff
@@ -436,6 +438,29 @@ impl orml_tokens::Trait for Runtime {
...
+    pub const ProxyDepositFactor: Balance = 1;
+    pub const MaxProxies: u16 = 4;
+    pub const MaxPending: u32 = 2;
+    pub const AnnouncementDepositBase: Balance = 1;
+    pub const AnnouncementDepositFactor: Balance = 1;
+}
+impl pallet_proxy::Trait for Runtime {
+    type Event = Event;
+    type Call = Call;
+    type Currency = orml_tokens::CurrencyAdapter<Runtime, ()>;
```
olanod, line 452: How does `()` implement the `Get<T::CurrencyId>` bound? We can set it to USDv since it's the main asset to be moving around, see how they do with the test https://github.com/open-web3-stack/open-runtime-module-library/blob/master/tokens/src/mock.rs#L123

olanod (commented):
Looks good, more restrictive proxy types can come later based on product needs

runtime/src/lib.rs
```diff
@@ -165,7 +166,8 @@ construct_runtime!(
         Sudo: pallet_sudo::{Call, Config<T>, Event<T>, Module, Storage},
         System: frame_system::{Call, Config, Event<T>, Module, Storage},
         Timestamp: pallet_timestamp::{Call, Inherent, Module, Storage},
-        Tokens: orml_tokens::{Config<T>, Event<T>, Module, Storage},
+        Tokens: orml_tokens::{Call, Config<T>, Event<T>, Module, Storage},
+        Proxy: pallet_proxy::{Call, Event<T>, Module, Storage}
```
olanod, line 170:
```suggestion
        Proxy: pallet_proxy::{Call, Event<T>, Module, Storage},
```

PR #63, opened 2021-03-08
feat:integrate orml-oracle pallet

stanly-johnson:
## Summary
Integrates the [orml-oracle](https://github.com/open-web3-stack/open-runtime-module-library/tree/master/oracle) pallet to allow price feeds on-chain. Currently the price feed can be simulated using [oracle-bot](https://github.com/stanly-johnson/orml-oracle-bot) but the plan is to use values from other onchain providers as they are ready.

Price of any asset can be fetch from other pallets using:

```rust
...
let vez_price = Oracle::get(&Asset::Fiat(Vez)), None);
...
```

#### Assumptions
- The baseCurrency is always `USD`, ie fetching the price of `VEZ` means the pair price of `VEZ/USD`
- Currently only the `sudo` account is permitted to add rates, will need to migrate this to a whitelist

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -253,6 +253,23 @@ impl vln_transfers::Config for Runtime {
     type Assets = Tokens;
 }
 
+parameter_types! {
+    pub const MinimumCount: u32 = 3;
+    pub const ExpiresIn: u32 = 600;
+    pub RootOperatorAccountId: AccountId = Sudo::key();
+}
+
+impl orml_oracle::Config for Runtime {
+    type Event = Event;
+    type OnNewData = ();
+    type CombineData = orml_oracle::DefaultCombineData<Runtime, MinimumCount, ExpiresIn>;
+    type Time = Timestamp;
+    type OracleKey = Asset;
+    type OracleValue = u64;
+    type RootOperatorAccountId = RootOperatorAccountId;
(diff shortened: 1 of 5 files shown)


runtime/src/lib.rs
```diff
@@ -253,6 +253,23 @@ impl vln_transfers::Config for Runtime {
...
+    pub RootOperatorAccountId: AccountId = Sudo::key();
+}
+
+impl orml_oracle::Config for Runtime {
+    type Event = Event;
+    type OnNewData = ();
+    type CombineData = orml_oracle::DefaultCombineData<Runtime, MinimumCount, ExpiresIn>;
+    type Time = Timestamp;
+    type OracleKey = Asset;
+    type OracleValue = u64;
```
olanod, line 268: Wonder if this will work in the future or if we'll need a fixed point type 🤔

olanod (approved):

runtime/src/lib.rs
stanly-johnson: I've changed it to [FixedU128](https://docs.rs/sp-runtime/2.0.0-rc4/sp_runtime/struct.FixedU128.html) to align with [Acala](https://github.com/AcalaNetwork/Acala/blob/2a445fe0b69ede82833acdba8c5213cf55579f82/modules/support/src/lib.rs#L25)

PR #69, opened 2021-03-15
feat:merge runtimes

### LINKED ISSUE #68: Merge runtimes
To reduce complexity of updating two separate runtimes whenever something changes we can have a single runtime that adds the parachain related code using compile time features.

stanly-johnson:
Merge `node` and `parachain` runtime

Resolve #68


runtime/src/lib.rs
```diff
@@ -46,8 +66,12 @@ pub use sp_runtime::{Perbill, Permill};
 pub type BlockNumber = u32;
 
 /// Alias to 512-bit hash when used in the context of a transaction signature on the chain.
+#[cfg(not(feature = "standalone"))]
 pub type Signature = MultiSignature;
 
+#[cfg(feature = "standalone")]
+pub type Signature = sr25519::Signature;
```
stanly-johnson, line 73: Hi @olanod, can I switch to `MultiSignature` for node aswell?

olanod: Sure, just check if any pallet complains or something. I think I left it that way to avoid some old code from complaining.

runtime/src/lib.rs
```diff
@@ -82,19 +101,19 @@ pub mod opaque {
...
+    pub struct SessionKeys {}
 }
 
 pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("VLN"),
-    impl_name: create_runtime_str!("vln-parachain"),
+    impl_name: create_runtime_str!("vln-runtime"),
     authoring_version: 1,
-    spec_version: 100,
+    spec_version: 1,
```
olanod, line 116: I remember reading in some issue somewhere that as convention the parachains where going to use the spec version starting at 100, we can check

runtime/src/lib.rs
```diff
@@ -106,7 +125,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 /// up by `pallet_aura` to implement `fn slot_duration()`.
 ///
 /// Change this to adjust the block time.
-pub const MILLISECS_PER_BLOCK: u64 = 12000;
+pub const MILLISECS_PER_BLOCK: u64 = 3000;
```
olanod, line 128: Have you tried this? As standalone chain I arbitrarily set this to 3s as I think we can achieve that target and with some fine tuning and perhaps stricter hardware requirements for our validators perhaps bring it down even more. But in parachain mode I think for the moment we have to stick to the block times of the relay chain. Is my understanding that this will change in the near future so we can go back to faster block production

runtime/src/lib.rs
```diff
@@ -236,31 +255,172 @@ impl pallet_sudo::Config for Runtime {
...
+            <Self::KeyOwnerProofSystem as KeyOwnerProofSystem<(KeyTypeId, GrandpaId)>>::Proof;
+        type KeyOwnerIdentification = <Self::KeyOwnerProofSystem as KeyOwnerProofSystem<(
+            KeyTypeId,
+            GrandpaId,
+        )>>::IdentificationTuple;
+        type HandleEquivocation = ();
+        type WeightInfo = ();
+    }
+
+    impl vln_foreign_asset::Config for Runtime {
```
olanod, line 331: Are our pallets required to be inside the feature gated block?

runtime/src/lib.rs
```diff
@@ -278,7 +438,6 @@ pub type SignedExtra = (
     frame_system::CheckTxVersion<Runtime>,
     frame_system::CheckGenesis<Runtime>,
     frame_system::CheckEra<Runtime>,
-    frame_system::CheckMortality<Runtime>,
```
olanod, line 281: The error I had in polkadot.js mentioned this check mortality thing, is it something that got removed from the template recently? Do you know what it means to not have it?

stanly-johnson: I'm not 100% sure, I changed it since all templates were using it, it[ appears to be a rename](https://github.com/paritytech/substrate/blob/c939ceba381b6313462d47334f775e128ea4e95d/frame/system/src/lib.rs#L125)

runtime/src/lib.rs
stanly-johnson: Not really, I thought will do that in the next PR along with integrating xcm, but no problem, I will add it here

runtime/src/lib.rs
stanly-johnson: Fixed in https://github.com/valibre-org/vln-node/pull/69/commits/b3c3956df5232948ca31e9e6e6c69c5d2326e6ec

runtime/src/lib.rs
stanly-johnson: Fixed in https://github.com/valibre-org/vln-node/pull/69/commits/6d72be08eb4d6d439473c3bc9349d8e364fa1ecc

runtime/src/lib.rs
stanly-johnson: Fixed in https://github.com/valibre-org/vln-node/pull/69/commits/8049eb6647f8f6f16a1fc3ddbb560c251cf9f530

olanod (approved):

olanod (approved):

PR #71, opened 2021-03-17
feat: integrate xcm and xtokens

stanly-johnson:
This PR integrates xcm and xtokens pallet to the parachain. 

Xtokens allows transfer of asset to other parachain that have xtokens integrated.
Xcm allows for transfer of assets between relaychain/parachains

Creating PR to `rococo-v1` branch since these features are subject to constant changes and not stable yet.


runtime/src/lib.rs
```diff
@@ -135,11 +158,11 @@ pub const DAYS: BlockNumber = HOURS * 24;
 // 1 in 4 blocks (on average, not counting collisions) will be primary babe blocks.
 pub const PRIMARY_PROBABILITY: (u64, u64) = (1, 4);
 
-// #[derive(codec::Encode, codec::Decode)]
-// pub enum XcmpMessage<XAccountId, XBalance> {
-//     /// Transfer tokens to the given account from the Parachain account.
-//     TransferToken(XAccountId, XBalance),
-// }
+#[derive(codec::Encode, codec::Decode)]
+pub enum XcmpMessage<XAccountId, XBalance> {
```
olanod, line 162: All looks good. This is the only bit that I wouldn't know what's its use, is it just a leftover from the template that was used as example?

olanod (approved):

PR #73, opened 2021-03-25
create instances of tokens pallet

stanly-johnson:
Related to [Instantiable Tokens](https://github.com/valibre-org/open-runtime-module-library/pull/2)

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -29,7 +29,8 @@ use sp_version::RuntimeVersion;
 mod proxy_type;
 use orml_traits::parameter_type_with_key;
 use proxy_type::ProxyType;
-use vln_primitives::Asset;
+use vln_primitives::{Asset, Collateral};
+use orml_tokens::CurrencyAdapter;
 
 #[cfg(feature = "standalone")]
 use standalone_use::*;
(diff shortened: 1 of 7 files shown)


stanly-johnson:
@olanod One side effect of having different instances of orml-tokens is multiple storage. Should I work on moving storage to a common external pallet or is this what we need? 🤔

runtime/src/lib.rs
```diff
@@ -397,10 +416,11 @@ macro_rules! construct_vln_runtime {
                     Sudo: pallet_sudo::{Module, Call, Storage, Config<T>, Event<T>},
 
                     // vln dependencies
-                    Tokens: orml_tokens::{Config<T>, Event<T>, Module, Storage},
-                    Proxy: pallet_proxy::{Call, Event<T>, Module, Storage},
+                    TokensGeneralInstance: orml_tokens::<Instance1>::{Config<T>, Event<T>, Module, Storage},
+                    TokensCollateralInstance: orml_tokens::<Instance2>::{Config<T>, Event<T>, Module, Storage},
```
olanod, line 420: Maybe can we just call it `Collateral`? 🙂 just a nitpick to be able to query the state with the VLN API doing something like `GET /collateral/total-issuance/ausd` instead of `/tokens-collateral-instance/total-issuance/ausd`

olanod: In the same sense we could keep calling the `TokensGeneralInstance` just `Tokens`, what do you think?

olanod (commented):
Separate storage per instance is expected, I don't see why it would affect us

olanod:
Should we close this one?

PR #74, opened 2021-03-30
feat: restrict mint to whitelist

stanly-johnson:
Restrict `attestation` call origin to Whitelist


stanly-johnson:
@olanod One caveat to current implementation is that the call to attestation should go through the collective pallet, like
`Whitelist::execute(ForeignAssets::attest(USDv, 100))`

Is this ok or do you want me to change it?

runtime/src/lib.rs
```diff
@@ -332,6 +348,34 @@ impl orml_oracle::Config for Runtime {
     type WeightInfo = ();
 }
 
+impl pallet_membership::Config for Runtime {
+	type Event = Event;
+	type AddOrigin = EnsureRoot<AccountId>;
+	type RemoveOrigin = EnsureRoot<AccountId>;
+	type SwapOrigin = EnsureRoot<AccountId>;
+	type ResetOrigin = EnsureRoot<AccountId>;
+	type PrimeOrigin = EnsureRoot<AccountId>;
```
stanly-johnson, line 357: Also how should we handle adding new members?

olanod:
> @olanod One caveat to current implementation is that the call to attestation should go through the collective pallet, like
> `Whitelist::execute(ForeignAssets::attest(USDv, 100))`
> 
> Is this ok or do you want me to change it?

ohh yeah is kind of ugly 🤔 what happens when you call the extrinsic directly, a bad origin error?(btw this reminds me I should add something the blockchain plugin to submit extrinsic through another module like `POST /_vln/foreign-assets/attest?via=sudo`) anyway I think conceptually we don't want to use the pallet collective as the attestation is not made by a group of individuals trying to agree on something but by independent individuals that happen to be allowed to perform some action, besides that pallet does more than we need. So I would change the generic `Whitelist` parameter of foreign assets to not be an origin but a generic type with the bound `frame_support::traits::Contains` which happens to be implemented by the `pallet-membership`, that way after doing the usual check of the origin being signed we just `ensure!(Whitelist::contains(origin))`

runtime/src/lib.rs
```diff
@@ -312,7 +328,7 @@ impl vln_human_swap::Config for Runtime {
 
 impl vln_transfers::Config for Runtime {
     type Event = Event;
-    type Assets = Tokens;
+    type Assets = Collateral;
```
olanod, line 331: I think this would still be be `Tokens`(where USDv is stored) as Collateral wont be transferred with this pallet, for now collateral stays with the person that minted/`attest`ed it.

olanod: Similarly the `Assets` of foreign-asset would rather be `Collateral` as we will `attest` things like `USDC` held externally

olanod (changes requested):
Awesome! looks good, just a minor change :)

olanod:
The ORML change is merged to our master, feel free to update this PR :)

PR #75, opened 2021-03-31
Implement similar asset transfer

stanly-johnson:
Implements basic transfer functionality where `from_currency == to_currency`


olanod (approved):
Should be a good start :)

PR #81, opened 2021-04-12
update to latest rococo

stanly-johnson:


runtime/src/lib.rs
```diff
@@ -549,13 +543,75 @@ mod parachain_impl {
...
+        X3(
+            Parent,
+            Parachain {
+                id: ParachainInfo::get().into(),
+            },
+            GeneralKey(id.encode()),
+        )
+    }
+
+    pub struct CurrencyIdConvert;
```
olanod, line 556: The only thing bothering me a bit is this block of the conversion logic, can we have it in a file next to the lib.rs?

olanod (approved):

PR #82, opened 2021-04-13
make foreigncurrencyid to network asset

### LINKED ISSUE #78: Better assets format
#76 introduces `ForeignCurrencyId` in addition to `Assets` Its better to merge/combine them to the extent possible that also works in Xcm environment.

stanly-johnson:
resolves #78


primitives/src/asset.rs
```diff
@@ -64,3 +67,23 @@ enum_with_aux_fns! {
         Vez = "VEZ",
```
olanod, line 67: Since the clippy restriction is already lifted, maybe make all currencies upper case?

olanod (approved):

PR #88, opened 2021-04-21
Human swap description

olanod:
Describe the flow of the swaps in more detail

### DIFF
--- a/pallets/human_swap/README.md
+++ b/pallets/human_swap/README.md
@@ -1,3 +1,36 @@
 # Human Swap Pallet
 
-Swaps that require human interaction(potentially automated on the human side), these humans(liquidity providers) are willing to exchange their own assets at a rate often higher than that of a configured price oracle, where one of the assets in the trading pair is a foreign(off-chain) asset, meaning that its completion depends on real world actions that when confirmed release the scrowed funds of the on-chain asset being traded.
+For swaps that require human interaction due to one of the trading pairs being a foreign asset(an asset that exists off-chain).  
+Users of this pallet specify a human they want to trade with based on the convenience of their rates, the party sending cryptocurrency
+gets its funds locked in a scrow and the other receives the information necesary to conduct the off-chain payment.
+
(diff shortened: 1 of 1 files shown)


pallets/human_swap/README.md
```diff
@@ -1,3 +1,36 @@
...
+
+![swap_flow](https://user-images.githubusercontent.com/1329925/115603536-90c2ed80-a2e0-11eb-810a-ebbe464d115a.png)
+
+This pallet tracks the state of the swaps which are updated based on actions form the user and the human involved. In the image each of the actions for either a `swap_in` or a `swap_out` changes the state of the overal swap which looks something like
+```rust
+struct Swap {
+  human: AccountId,
+  type: SwapType,
+  rate: PairRate,
+  amount: Balance,
```
stanly-johnson, line 15: Shouldn't we be storing the swap-recipient also? especially in cash-out we need this account to reserve Usdv and then transfer to LP?

stanly-johnson (approved):
This looks good. The only doubt I have is that in both cash-in and cash-out, there is no real incentive to call step 4. So maybe step4 should be automatically triggered on step3? If this is how you meant it then its fine :)

pallets/human_swap/README.md
olanod: I was thinking we already have that info since it makes sense to have the swap information indexed by the user account and swap id as keys of the storage, i.e. `(UserAccountId, SwapId) -> Swap` so we can query the swaps of a user or a specific one. Actually the `SwapId` there can be just a nonce, like the one you would get from `System::account_nonce` and what we call `SwapId` be the combination of both the user account and the nonce.

olanod:
Good point about the `release` command. It's true that there is not incentive since the person that received the money is good to go but since the funds are still locked there the party pending payment can "dispute" so a judge can come and verify the proofs to release the locked funds. 
Something extra to explore that can come at the cost of poor UX is to game the process requiring the party that locks the funds to lock an extra percentage that is held as guarantee and given back after the release.

stanly-johnson (approved):

PR #90, opened 2021-04-24
xcm local testing scripts

stanly-johnson:
Bunch of scripts that will help with testing setting-up, onboarding and transfer assets on xcm


olanod:
This kind of scripts seem useful but I'm hesitant to bring the node dependency to the repo 🤔 Do you think this could fit in a new repo like `vln-toolbox` that might evolve into a feature-full cli at some point in time?

stanly-johnson:
Ya makes sense, I'll create a new repo for these scripts

PR #95, opened 2021-05-07
Escrow pallet

stanly-johnson:
This pallet will allow users to create an escrow onchain for a selected recipent. Also implements `EscrowHandler` trait that allows other modules to utilise the escrow functions.

### DIFF
--- a/pallets/escrow/README.md
+++ b/pallets/escrow/README.md
@@ -1,0 +1,3 @@
+# Escrow Pallet
+
+This pallet will allow users to create an escrow onchain for any peer-to-peer transaction.
--- a/pallets/escrow/README.md
+++ b/pallets/escrow/README.md
@@ -1,6 +1,6 @@
 # Escrow Pallet
 
-This pallet will allow users to create an escrow onchain for a selected recipent. The pallet can be used to compliment the human swap pallet, to prove the existence of an escrow before the actual swap needs to be processed.
+This pallet will allow users to create an escrow onchain for a selected recipent. The pallet can be used to compliment the human swap pallet, to prove the existence of an escrow before the actual swap needs to be processed. The escrow pallet does not store the history of all the escrow transactions created by the user, it only cares about the current active escrow of the user/recipent.
 
 ## Terminology
 
--- a/pallets/escrow/README.md
+++ b/pallets/escrow/README.md
@@ -1,3 +1,31 @@
 # Escrow Pallet
 
-This pallet will allow users to create an escrow onchain for any peer-to-peer transaction.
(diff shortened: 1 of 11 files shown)


pallets/escrow/src/lib.rs
```diff
@@ -0,0 +1,171 @@
...
+    #[pallet::getter(fn rates)]
+    /// Escrows created by a user, this method of storageDoubleMap is chosen since there is no usecase for
+    /// listing escrows by provider/currency. The escrow will only be referenced by the creator in
+    /// any transaction of interest.
+    pub(super) type Escrow<T: Config> = StorageDoubleMap<
+        _,
+        Twox64Concat,
+        T::AccountId,
+        Twox64Concat,
+        EscrowId,
```
olanod, line 49: I think we should enforce people having only one escrow with any other user at a time, I don't see situations where a user might have more than one open trade with the same person(they can cancel and recreate the escrow with a different amount) in turn this restriction can help with automation and mitigate the cases where the user has more than one transfer in their account from the same recipient with the same amount.
All that just to say that the EscrowId can be just the recipient Account ID, so releasing an escrow for example just needs to know the recipient account.

pallets/escrow/src/lib.rs
```diff
@@ -0,0 +1,171 @@
...
+        EscrowDetail<T::AccountId, AssetIdOf<T>, BalanceOf<T>>,
+    >;
+
+    /// Current escrow index for a user
+    #[pallet::storage]
+    #[pallet::getter(fn swap_index)]
+    pub(super) type EscrowIndex<T: Config> =
```
olanod, line 56: This won't likely be needed if with what I mentioned in the other comment :)

olanod (changes requested):

olanod (commented):
Looking good :) a couple of questions though

pallets/escrow/README.md
```diff
@@ -1,6 +1,6 @@
 # Escrow Pallet
 
-This pallet will allow users to create an escrow onchain for a selected recipent. The pallet can be used to compliment the human swap pallet, to prove the existence of an escrow before the actual swap needs to be processed.
+This pallet will allow users to create an escrow onchain for a selected recipent. The pallet can be used to compliment the human swap pallet, to prove the existence of an escrow before the actual swap needs to be processed. The escrow pallet does not store the history of all the escrow transactions created by the user, it only cares about the current active escrow of the user/recipent.
```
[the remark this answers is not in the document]
olanod: 👌🏽 got it

PR #96, opened 2021-05-11
cancel escrow fn

stanly-johnson:
`cancel_escrow` extrinsic for escrow pallet

### DIFF
--- a/pallets/escrow/src/lib.rs
+++ b/pallets/escrow/src/lib.rs
@@ -83,7 +83,7 @@ pub mod pallet {
         /// The only action is to store the details of this escrow in storage and reserve
         /// the specified amount.
         #[pallet::weight(10_000 + T::DbWeight::get().writes(1))]
-        pub fn create_escrow(
+        pub fn create(
             origin: OriginFor<T>,
             recipent: T::AccountId,
             asset: AssetIdOf<T>,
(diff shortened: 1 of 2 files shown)


stanly-johnson:
Thinking about the dispute mechanism we plan to implement, if a user creates an escrow, receives the asset off chain and then proceeds to cancel the escrow, then the dispute mechanism can't do much. Maybe we can add a delay of `x blocks` before the user can cancel, to give the recipient time to raise a dispute, but this could be an annoyance in case the user just wants to update the amount with the same provider. Maybe adding an update escrow (with limitations) and delaying cancellation will help? But there is also the case of wanting to switch providers but now the money is locked up🤔 wdyt?

pallets/escrow/src/lib.rs
```diff
@@ -56,10 +56,12 @@ pub mod pallet {
     #[pallet::metadata(T::AccountId = "AccountId")]
     #[pallet::generate_deposit(pub(super) fn deposit_event)]
     pub enum Event<T: Config> {
-        /// Rate has been updated
+        /// A new escrow has been created
         EscrowCreated(T::AccountId, AssetIdOf<T>, BalanceOf<T>),
-        /// Rates have been removed by LP
+        /// Escrow amount released to the recipent
```
olanod, line 61: Good catch, didn't see it last time 😅

olanod:
> Thinking about the dispute mechanism we plan to implement, if a user creates an escrow, receives the asset off chain and then proceeds to cancel the escrow, then the dispute mechanism can't do much. Maybe we can add a delay of `x blocks` before the user can cancel, to give the recipient time to raise a dispute, but this could be an annoyance in case the user just wants to update the amount with the same provider. Maybe adding an `update_escrow` (with limitations, ie. only allow increase or decrease by x%) and delaying cancellation will help? But there is also the case of wanting to switch providers but now the money is locked upthinking wdyt?

I overlooked the logic and didn't notice you expect the user to cancel. The user(creator of the escrow) can not call the cancel, only the recipient in this case can cancel or you face the problem you mention. With the dispute system the third-party judging the transaction can also cancel or release. And yes as a follow-up PR we can add a configuration for the pallet to specify the delay in blocks that a transaction would be automatically cancelled for those cases when the recipient is unresponsive

olanod (approved):

PR #98, opened 2021-05-11
Allow setting fixed rates

### LINKED ISSUE #97: Allow rate providers to update fixed rates
Current implementation of the rate provider pallet expects the providers to upload rate premium and not the actual rate, we could support both `fixed_rate` and `rate_premium` and let the provider choose the format.

Current:
```rust
pub struct RateDetail {
    pub rate: Permill,
}
```

Change to:
```rust
pub enum RateFormat {
 Fixed(FixedU128),
 Premium(Permill)
}

pub struct RateDetail {
    pub rate: RateFormat,
}
```

stanly-johnson:
Resolves #97


olanod (approved):
Nice :)

PR #109, opened 2021-06-03
Ci automate release

### LINKED ISSUE #108: Automate releases
Using GitHub actions, create a new release when a new tagged commit is pushed to master

stanly-johnson:
Github Action to create a new release when a new tag is pushed. [Example](https://github.com/stanly-johnson/vln-node/releases/tag/v0.0.3)
resolves #108

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,69 @@
+name: Create Release
+
+on:
+  push:
+    branches:
+      - automate-release
+    tags:
+      - 'v*.*.*'
+
+jobs:
+  build:
+    # name: Build project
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
+name: Create Release
+
+on:
+  push:
+    tags:
+      - 'v*.*.*'
```
olanod, line 6: Maybe we can start tagging our releases with the runtime version that we'll need to update on every change?

olanod (dismissed):

olanod (approved):

PR #112, opened 2021-06-05
Switch to assets pallet

### LINKED ISSUE #84: Replace tokens pallet instance with assets
Related [substrate/8483](https://github.com/paritytech/substrate/pull/8483)
Depends on [substrate/8453](https://github.com/paritytech/substrate/issues/8453) specifically `fungibles::Lockable`

stanly-johnson:
Use pallet-assets to generate tokens

- removes orml-tokens
- adds pallet-balances, pallet-transaction-payment, pallet-assets
- convert escrow pallet to use fungible::* traits

resolves #84


runtime/Cargo.toml
```diff
@@ -11,66 +11,68 @@ version = '2.0.0'
 targets = ['x86_64-unknown-linux-gnu']
 
 [build-dependencies]
-substrate-wasm-builder = "4.0.0"
+substrate-wasm-builder = '4.0.0'
```
olanod, line 14: Are the single vs double quotes something your editor does or from the template? I don't mind either but would be nice to avoid tools changing things back and forth making harder to review dependency changes.

runtime/src/lib.rs
```diff
@@ -407,12 +365,64 @@ impl vln_rate_provider::Config for Runtime {
...
+    pub const AssetDeposit: Balance = UNITS; // 1 UNIT deposit to create asset
+    pub const ApprovalDeposit: Balance = UNITS;
+    pub const StringLimit: u32 = 50;
+    /// Key = 32 bytes, Value = 36 bytes (32+1+1+1+1)
+    // https://github.com/paritytech/substrate/blob/069917b/frame/assets/src/lib.rs#L257L271
+    pub const MetadataDepositBase: Balance = UNITS;
+    pub const MetadataDepositPerByte: Balance = UNITS;
+}
+
+impl pallet_assets::Config for Runtime {
```
olanod, line 404: Can we already have a second instance of assets for `fiat`?

olanod (changes requested):
About the escrow I think we can still keep it around already changing the implementation to depend on one of the fungible traits that has the minimum functionality we might need(e.g. https://github.com/paritytech/substrate/blob/master/frame/support/src/traits/tokens/fungibles.rs#L153) and then instantiate it with a dummy(`e.g. `Asset = ()` if `()` implements the trait)

olanod (approved):
😃👌🏽

Issue #111, opened 2021-06-04
Uniques pallet and NFTs to represent goods or services in the real world


olanod:
In a similar way to how people can tokenize their fiat currency with the assets pallet to make the chain aware of a user's willingness to trade said fiat for a crypto asset, the same concept can be applied to any non-fungible thing that exists in the real world that a user wants to trade for some cryptocurrency. For example, as a seller I might want to get rid of some old shoes so I mint an [NFT](https://github.com/paritytech/substrate/tree/master/frame/uniques) in VLN for which I set a price and medium of payment in the rates pallet(is it still a good name? or prices/trades/deals pallet?), client applications showing a market place of used shoes can query the state of the chain to get the list of all prices for the NFT class for used shoes, a buyer will select the shoes and start a trade with the seller in an encrypted matrix room putting the funds in an escrow in the exact same way as with a fiat trade, once the shoes arrive the escrow is released.

olanod:
No longer applicable as Matrix home servers will keep track of the listings of products and services

PR #115, opened 2021-06-10
Add uniques pallet

stanly-johnson:
related #111


stanly-johnson:
@olanod Is anything else needed here? I assume from the PoV of escrow the NFT is just like the fiat, we rely on the recipient to validate that the transfer happened. We can definitely improve this, maybe an auction model for NFTs 🤔

olanod (approved):

PR #117, opened 2021-06-28
Update to latest xtokens

### LINKED ISSUE #85: Update to xcm revamp
Depends on https://github.com/paritytech/polkadot/issues/2841

stanly-johnson:
Reintegrate xcm asset transfer after xcm revamp changes.
Use pallet-assets in place of xtokens

Resolves #85


runtime/src/lib.rs
```diff
@@ -532,139 +554,159 @@ mod parachain_impl {
...
+    }
+
+    impl cumulus_pallet_dmp_queue::Config for Runtime {
+        type Event = Event;
+        type XcmExecutor = XcmExecutor<XcmConfig>;
+        type ExecuteOverweightOrigin = EnsureRoot<AccountId>;
+    }
+
+    pub type LocalAssetTransactor = MultiCurrencyAdapter<
+        NetworkAsset,
```
stanly-johnson, line 677: For time being, I have added a new instance of orml-tokens to allow for token transfer to get around the implementation requirement of `orml_traits::MultiCurrency`. Will think of a better solution, either replacing the MultiCurrencyAdapter implementation or supporting pallet_assets in xtokens 🤔

olanod: `xtokens` doesn't depend on `MultiCurrency`, seems we only need it for this [MultiCurrencyAdapter](https://github.com/open-web3-stack/open-runtime-module-library/blob/master/xcm-support/src/currency_adapter.rs#L44) from `xcm-support`. The code there is simple as it's just a small utility implementing the [`TransactAsset`](https://github.com/paritytech/polkadot/blob/master/xcm/xcm-executor/src/traits/transact_asset.rs#L28) trait. I think is worth creating an `AssetAdapter`/`AssetTransactor` to replace this currency adapter, later we can propose it to the ORML repo and that way we reduce our dependencies.

olanod (changes requested):

runtime/src/lib.rs
stanly-johnson: Removed orml-tokens and modified MultiCurrencyAdapter with `Mutate` trait, but not using it to mint/burn right now, that requires more changes wrt CurrencyId/AssetId. For now every xcm deposit will be to UnknownTokens.

stanly-johnson: #121 Will work on this next, don't want to block the demo

olanod:
Still WIP?

runtime/src/lib.rs
```diff
@@ -712,11 +736,12 @@ construct_vln_runtime! {
...
-    // UnknownTokens: orml_unknown_tokens::{Pallet, Storage, Event},
-    // PolkadotXcm: pallet_xcm::{Pallet, Call, Event<T>, Origin},
-    // CumulusXcm: cumulus_pallet_xcm::{Pallet, Origin},
-    // XcmpQueue: cumulus_pallet_xcmp_queue::{Pallet, Call, Storage, Event<T>},
+    XcmpQueue: cumulus_pallet_xcmp_queue::{Pallet, Call, Storage, Event<T>},
+    PolkadotXcm: pallet_xcm::{Pallet, Call, Event<T>, Origin},
+    CumulusXcm: cumulus_pallet_xcm::{Pallet, Event<T>, Origin},
+    DmpQueue: cumulus_pallet_dmp_queue::{Pallet, Call, Storage, Event<T>},
+    XTokens: orml_xtokens::{Pallet, Storage, Call, Event<T>},
+    UnknownTokens: orml_unknown_tokens::{Pallet, Storage, Event},
```
olanod, line 744:
```suggestion
    UnknownAssets: orml_unknown_tokens::{Pallet, Storage, Event},
```

runtime/src/currency_id_convert.rs
```diff
@@ -1,64 +1,64 @@
...
-// impl Convert<NetworkAsset, Option<MultiLocation>> for CurrencyIdConvert {
-//     fn convert(id: NetworkAsset) -> Option<MultiLocation> {
-//         match id {
-//             NetworkAsset::DOT => Some(X1(Parent)),
-//             NetworkAsset::ACA | NetworkAsset::AUSD => Some(native_currency_location(id)),
-//         }
-//     }
-// }
+pub struct CurrencyIdConvert;
+impl Convert<u32, Option<MultiLocation>> for CurrencyIdConvert {
```
olanod, line 14: I like it how before was more explicit using the `NetwotkAsset` type instead of `u32` making it more explicit. Any special reason to have u32? I think even an alias would be better.

runtime/src/transact_asset.rs
```diff
@@ -0,0 +1,120 @@
...
+        }
+    }
+}
+
+/// The `TransactAsset` implementation, to handle `MultiAsset` deposit/withdraw.
+/// Note that teleport related functions are unimplemented.
+///
+/// If the asset is known, deposit/withdraw will be handled by `MultiCurrency`,
+/// else by `UnknownAsset` if unknown.
+pub struct MultiCurrencyAdapter<
```
olanod, line 48: Don't love the way this is sooo generic having to deal with all the type gymnastics but not sure if it's avoidable.
Also since we don't use `MultiCurrency` in our system perhaps a better name could be something like `AssetsAdapter`? or `NetworkAssetsAdapter`?

olanod (changes requested):
Awesome that we can bring back cross-chain functionality 🎉

runtime/Cargo.toml
```diff
@@ -54,23 +54,25 @@ cumulus-primitives-core = {git = 'https://github.com/paritytech/cumulus', defaul
 parachain-info = {git = 'https://github.com/paritytech/cumulus', default-features = false, branch = 'polkadot-v0.9.7'}
 cumulus-pallet-aura-ext = { git = 'https://github.com/paritytech/cumulus', branch = 'polkadot-v0.9.7', default-features = false }
 cumulus-primitives-timestamp  = { git = 'https://github.com/paritytech/cumulus', branch = 'polkadot-v0.9.7', default-features = false }
-# xcm = {git = 'https://github.com/paritytech/polkadot', default-features = false, branch = 'polkadot-v0.9.7'}
-# xcm-builder = {git = 'https://github.com/paritytech/polkadot', default-features = false, branch = 'release-v0.9.7'}
-# xcm-executor = {git = 'https://github.com/paritytech/polkadot', default-features = false, branch = 'release-v0.9.7'}
+xcm = {git = 'https://github.com/paritytech/polkadot', default-features = false, branch = 'release-v0.9.7'}
```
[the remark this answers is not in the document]
stanly-johnson: orml is still on 0.9.7, we have to wait for them to update it

runtime/src/currency_id_convert.rs
stanly-johnson: The u32 comes from [here](https://github.com/valibre-org/vln-node/blob/38c5570546b65da72e1007dd159a05ccf0c1d7ab/runtime/src/lib.rs#L416), the type in pallet_assets and here should match, do you want me to type alias u32 here?

olanod: mmm yeah I think helps readability

olanod (approved):

PR #124, opened 2021-07-01
Docker compose for local testnet 

olanod:
Adds a full local setup of the relay chain validators and two parachains(vln & karura) using docker containers for better portability and quicker deployment.


olanod:
@stanly-johnson let me know if this set up works in your machine. Since I'm on Linux using rootless podman containers(+podman-compose) there might be differences between my setup and the usual docker setup.
Additionally the second commit(that could well be a separate PR) is for setting the para-id to something fixed to not mess in the future with those parameters.

stanly-johnson (approved):
Awesome 🚀  Works for me locally

PR #126, opened 2021-07-07
Comments

jeluard:


pallets/rate_provider/src/lib.rs
```diff
@@ -100,6 +105,7 @@ pub mod pallet {
             medium: PaymentMethod,
             rate: RateDetailOf<T>,
         ) -> DispatchResultWithPostInfo {
+// SBP1 No need to use DispatchResultWithPostInfo if always returning default
```
olanod, line 108: 👍🏽

pallets/rate_provider/src/lib.rs
```diff
@@ -134,6 +140,7 @@ pub mod pallet {
                     Ok(())
                 },
             )?;
+// SBP1 What if removing when no rate is provided? Shall the event be updated?
```
olanod, line 143: If the triplet doesn't exist we shouldn't emit an event.

olanod (commented):
Thanks for the feedback @jeluard, we'll ping you in a following PR to address some of the points of the review :)

PR #145, opened 2021-10-29
Renamed all instances of escrow into payment

### LINKED ISSUE #140: Rename escrow pallet to payments
Not everyone immediately understands what an escrow is and how is supposed to work unlike "payment" which is universal, besides escrows imply there is a third party holding funds in custody but in our case we just freeze the funds of the creator and proceed to transfer when releasing the escrow. Another advantage of the rename is to implement a "payment request" that is simpler to understand than "escrow request".  

To make the transaction align better with the concept of a payment, when creating it we should already transfer the funds and freeze them on the receiving end, this enables more complex future use cases like using the newly received locked funds as a promise for another transaction(not waiting for the first one to finish).

KenjiPcx:
Closes #140


olanod (approved):
Made a change to the readme and ran `cargo fmt` to keep the CI pipeline happy

PR #151, opened 2021-12-01
[feat] Update payment flow

### LINKED ISSUE #147: Payments are sent immediately to the destination
Currently payments are reserved in the origin account and only transferred to the destination after the release call is made. We should change it so the transfer of funds is made immediately when the payment is created but the funds are reserved on arrival.

This can help the seller prove to another party that it has funds that are just pending a release. i.e. a seller could borrow liquidity from other protocol using its future funds as guarantee.

stanly-johnson:
### Update payment flow

Existing Flow :
In the current payment flow when a payment is created the amount is reserved on the sender wallet and when the “release” call is made the actual transfer to recipient wallet is executed.

New Flow:
When a new payment is created, the amount is immediately transferred from sender’s wallet to recipient’s wallet. The amount is then reserved in the recipient wallet until the “release” call is made. The `incentive_amount` will be reserved in the payment creators account

Closes #147
Requires #149


olanod (approved):
Nice, looks good to me!

Issue #128, opened 2021-09-04
Permisionless version of escrow pallet


stanly-johnson:
One of the things that could make escrow pallet more "permissionless" would be to have a council in place of a whitelist judge. In case of a dispute the council can vote on wheter to release/return the escrow. Members of this council have to lock in tokens to be part of the council, which should make them invested in the success of the network. This of course assumes a token based model.

What do you think? @olanod

olanod:
**For disputes**: From the point of view of the escrow pallet, it shouldn't care about councils or whatever, each escrow when created gets assigned a designated origin that will have the permissions to resolve the conflict, this origin can be stored in the `EscrowDetail` and we can get it from a configurable generic provider(e.g. a `DisputeResolver::get_origin()`). For now we just provide an implementation that returns the Sudo origin and later we create a more dynamic version that can get the the canonical account of community which could be ruled by a council or whatever the community decides.

**For the escrow creator**:  In https://github.com/virto-network/virto-node/issues/141 we would tackle how to incentivize the release of the escrow, there wouldn't be anything else to make it more "permissionless" other than the usual fees, if its necessary in the future we can add a more complex fee system that allows charging percentages based on the payment amount or something but it wouldn't be necessary for now.

PR #152, opened 2021-12-02
[feat] Dispute resolver trait

### LINKED ISSUE #143: Payment(escrow) DisputeResolver
Payment/Escrow details should also store a `resolver_account` which is set at the time of creation of the payment that we get from a `DisputeResolver::get_origin`. When resolving, we just check if the `resolver_account` matches the calling origin.

With the initial implementation let's create a `DisputeResolver` that just returns the Sudo origin.

stanly-johnson:
This PR creates a new `resolver_account` field in PaymentDetail struct. This allows every new payment to select a resolver account if desired, if a resolver is not selected by the payment creator the `DisputeResolver` trait is used to determine a resolver for the payment.

Currently the `DisputeResolver` trait simply returns the root account, this can be improved in future to add more complex resolver selection logic.

Closes #143, #128

### DIFF
--- a/pallets/payment/src/lib.rs
+++ b/pallets/payment/src/lib.rs
@@ -91,11 +91,10 @@ pub mod pallet {
 			recipient: T::AccountId,
 			asset: AssetIdOf<T>,
 			amount: BalanceOf<T>,
-			resolver: Option<T::AccountId>,
 		) -> DispatchResultWithPostInfo {
 			let who = ensure_signed(origin)?;
 			<Self as PaymentHandler<T::AccountId, AssetIdOf<T>, BalanceOf<T>>>::create_payment(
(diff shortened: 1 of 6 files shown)


pallets/payment/src/lib.rs
```diff
@@ -93,10 +91,11 @@ pub mod pallet {
 			recipient: T::AccountId,
 			asset: AssetIdOf<T>,
 			amount: BalanceOf<T>,
+			resolver: Option<T::AccountId>,
```
olanod, line 94: Nope, this parameter should go away. I don't think neither the person paying or getting payed should be able to influence in anyway what account has the permission to resolve the disputes since you could select a friend that is biased to vote in your favor.

olanod (changes requested):

Issue #130, opened 2021-09-14
Tracking issue : W3F Grants Milestone 1


stanly-johnson:
Refer : https://github.com/w3f/Grants-Program/blob/master/applications/plip.md

- [x] Switch from assets pallet to tokens pallet (#129)
- [x] Update xcm and xtokens support (#132)
- [x] Add benchmarks for escrow pallet (#155)
~- [ ] Add benchmarks for rates pallet~ (removed in #131)
- [ ] Launch public testnet and update README on how to create local version / access testnet

Issue #103, opened 2021-05-22
benchmarks for vln pallets


stanly-johnson:
How about benchmarks for vln custom pallets before going for parachain registration on Kusama?

- [x] payment (#155)
- [ ] communities

olanod:
This is definitely a must do before Kusama for sure. Although first we probably will spend some more time in the testnet world polishing the core functionality.

stanly-johnson:
I think the rates pallet looks good from a permisionless perspective. Adding weights to it should make it good to go for web3 grant. Wdyt @olanod ?

PR #155, opened 2021-12-08
[feat] Benchmarks for payments pallet

stanly-johnson:
Related to #103, #130


pallets/payment/src/benchmarking.rs
```diff
@@ -0,0 +1,76 @@
...
+use frame_system::RawOrigin;
+use orml_traits::MultiCurrency;
+use virto_primitives::PaymentState;
+
+const SEED: u32 = 0;
+
+fn setup_caller<T: Config>() -> T::AccountId {
+	let caller = whitelisted_caller();
+	let value: BalanceOf<T> = 10u32.into();
+	let asset: AssetIdOf<T> = "test".into(); // TODO : figure out way to convert to type of CurrencyId of orml-tokens
```
stanly-johnson, line 14: @olanod Not sure how to solve this 🤔 [CurrencyId trait](https://github.com/open-web3-stack/open-runtime-module-library/blob/17a791edf431d7d7aee1ea3dfaeeb7bc21944301/tokens/src/lib.rs#L188) is not very helpful. Acala seems to be [tightly coupling the CurrencyId](https://github.com/AcalaNetwork/Acala/blob/b2054d585665478422e9d14c0c8620dc832f708d/modules/homa-lite/src/lib.rs#L75), looks like a bad design for us with multiple instances.

olanod: Still not sure what the problem is, does it fail with `"test"`? if so, maybe the string literal simply doesn't implement the required traits. What about using one of the `Asset` variants?

stanly-johnson: The problem is I'm not sure how to convert any data to the "type" CurrencyId. When I use Asset, the error is 

```
note: expected associated type `<<T as pallet::Config>::Asset as MultiCurrency<<T as frame_system::Config>::AccountId>>::CurrencyId`
                         found enum `Asset`
```

But the [CurrencyId](https://github.com/open-web3-stack/open-runtime-module-library/blob/17a791edf431d7d7aee1ea3dfaeeb7bc21944301/tokens/src/lib.rs#L188) does not require anything that allows conversion. So the challenge is to convince the compiler to accept our mock data as CurrencyId 🤔

olanod:
Sorry I took long to take a look, there was this `where_clause` thing in the `benchmarks!` macro that allows specifying the trait bounds, didn't adapt the whole code since I'm on a rush but there's the idea. Also I was getting an unrelated error with a dependency of polkadot-runtime(collective pallet) not implementing something, perhaps I updated something I shouldn't

stanly-johnson:
> Sorry I took long to take a look, there was this `where_clause` thing in the `benchmarks!` macro that allows specifying the trait bounds, didn't adapt the whole code since I'm on a rush but there's the idea. Also I was getting an unrelated error with a dependency of polkadot-runtime(collective pallet) not implementing something, perhaps I updated something I shouldn't

Thanks! I didn't know we could use where_clause that way 💯 The error with collective pallet appears in the template and cumulus repo as well. I've created an issue, will fix here once that is resolved

pallets/payment/src/benchmarking.rs
```diff
@@ -0,0 +1,69 @@
...
+benchmarks! {
+	where_clause { where T::Asset: MultiCurrency<
+		<T as frame_system::Config>::AccountId,
+		CurrencyId = u32, Balance = u32
+	>
+}
+	// create a new payment succesfully
+	create {
+		let caller = whitelisted_caller();
+		let _ = T::Asset::deposit(1u32, &caller, 100u32);
```
olanod, line 28:
```suggestion
		let _ = T::Asset::deposit(TEST_ASSET, &caller, 100u32);
```
Maybe I'd prefer this to be a constant or something that describes better what it is since with a `1` is not very clear we are talking about an asset and could look like a quantity. Also, although it probably won't make a big difference, it might give more realistic results to use the actual asset type or something similar to get the benchmarks closer to the real thing, it might be that internally we serialize the asset parameter, store it or do something with it where the `u32` might produce different results that the enum we usually use(e.g. a simple enum in SCALE takes 1 byte vs 4 of the u32).

olanod (approved):

PR #161, opened 2021-12-15
[feat] Fee Handler

### LINKED ISSUE #158: apply_fees extrinsic
`apply_fees` that returns the amount that is actually sent to the merchant, by default it does nothing and returns the full amount but in our runtime we can configure a different handler that checks what market places the merchant is registered to and applies the discount of that marketplace(e.g. swap cash might take a 1% of commission as it allows people listing their fiat trades in its infrastructure(matrix room))

stanly-johnson:
resolves #158 

ToDo

- [x] Add more docs
- [x] Improve tests


stanly-johnson:
@olanod I created a simple version of the pallet to get started. The next stage would be converting 
`fn apply_fees(from: &Account, to: &Account, asset: Asset, amount: Amount) -> Vec<(Account, Amount)>;`

And within the `release` function to loop through all the recipients. Is that correct? 
Although I think this basic version is good enough for majority of cases where we will have only one recipient of fees. The drawback of the advanced approach is that the realease extrinsic ends up being expensive for the user, even though the actual recipient count might be low,  since we use the worst case in benchmark to calculate the weights which means a user always pays the transaction fee equivalent to transfer to our MAX_RECIPIENTS always. Wdyt? Should we go ahead with the `Vec<>` approach? 🤔

olanod:
About the worst case scenario, it's worth doing the extra work of calculating the weight dynamically or doing a weight correction https://docs.substrate.io/v3/runtime/weights-and-fees/#dynamic-weights

pallets/payment/src/types.rs
```diff
@@ -59,3 +59,7 @@ pub trait DisputeResolver<Account> {
 	/// Get a DisputeResolver (Judge) account
 	fn get_origin() -> Account;
 }
+
+pub trait FeeHandler<Account, Asset, Amount> {
+	fn apply_fees(from: &Account, to: &Account, asset: Asset, amount: Amount) -> Amount;
```
olanod, line 64: I'd say let's make this function return already the account that is receiving the fee instead of the pallet constant, later we can introduce the Vec thing to support multiple receivers.
I'm also hesitant about using an amount since there's more checks to make(e.g. can we withdraw that amount?) and merchants will usually just care about configurign a percentage(`-> (Account, Permill)`?).
Finally, to keep the trait simpler, at least at the begging, I think we can also ditch the `Asset` and `Amount` dependency since, at least for Virto, we only need to know the `from` and `to` to determine who gets the fee(e.g. local community or marketplace)

olanod: Let's also create a public dummy implementation of the trait for a generic `Account` that returns the account with a 0 fee and we use it in our runtime config :)

pallets/payment/src/lib.rs
```diff
@@ -233,7 +240,16 @@ pub mod pallet {
 					T::Asset::unreserve(payment.asset, &from, payment.incentive_amount);
 					// unreserve the amount to the recipent
 					T::Asset::unreserve(payment.asset, &to, payment.amount);
-
+					// calculate fee charged for transaction
```
olanod, line 243: I think it's better for the fee to be calculated at the moment of creation of the payment and not when released to avoid "surprises". As a customer when you pay for an article you want to know well from the beginning what you are paying, if we do this in the release stage is hard to predict what the fee will be. We can calculate at the moment of creation and store the amount to be charged in the payment detail in a `fees` field, then we transfer on release. Probably also wise to reserve the amount to make sure we can pay the marketplace(what happens if the account has no funds from the moment the payment is created and released? is payment cancelled?), named reserves will probably help making that bookkeeping clearer 🤔

olanod (changes requested):

pallets/payment/src/types.rs
stanly-johnson: Created #162 to track multiple recipients improvement

pallets/payment/src/lib.rs
```diff
@@ -240,15 +240,12 @@ pub mod pallet {
...
-					// calculate fee charged for transaction
-					let fee_amount =
-						T::FeeHandler::apply_fees(&from, &to, payment.asset, payment.amount);
 					// transfer fee amount to marketplace
 					T::Asset::transfer(
 						payment.asset,
-						&from,
-						&T::FeeRecipientAccount::get(),
-						fee_amount,
+						&from,                 // fee is paid by payment creator
```
olanod, line 246: aren't we missing the `unreserve` of the `fee_amount` so it doesn't fail if there are not enough funds?

pallets/payment/src/tests.rs
```diff
@@ -27,7 +27,8 @@ fn test_create_payment_works() {
 				amount: 20,
 				incentive_amount: 2,
 				state: PaymentState::Created,
-				resolver_account: RESOLVER_ACCOUNT
+				resolver_account: RESOLVER_ACCOUNT,
+				fee_detail: (FEE_RECIPIENT_ACCOUNT, 0)
```
olanod, line 31: Perhaps it's worth having a test case for an implementation of the fee handler that actually returns a fee so we can catch issues like the unreserve above ;)

olanod (changes requested):
Looks good, just a small thing :)

pallets/payment/src/tests.rs
stanly-johnson: The `test_charging_fee_payment_works` already exists but it was asserting wrong results 😅 I've fixed that and also added `test_charging_fee_payment_works_when_canceled`

olanod (approved):

PR #163, opened 2021-12-21
[feat] pay_with_detail extrinsic

### LINKED ISSUE #157: pay_with_detail extrinsic
We need a way for people to give some context to the payment for some use cases. To not create any form of coupling with other components it should be very generic, just allow to optionally provide some extra(limited) amount of bytes like a remark. We can rename create to pay and then have an extra call that is `pay_with_remark` or `pay_with_detail` where the user provides those extra arbitrary bytes.

In Virto that system can be used to incentivize people buying things via a marketplace instead of p2p if for example the payments without a remark have a higher fee that goes to the communities of the buyer/seller but if the remark with some data that identifies the marketplace is provided then the discount would be smaller and goes to the marketplace.

### LINKED ISSUE #158: apply_fees extrinsic

stanly-johnson:
resolves #157


pallets/payment/src/lib.rs
```diff
@@ -91,15 +96,38 @@ pub mod pallet {
...
+		#[pallet::weight(10_000 + T::DbWeight::get().writes(1))]
+		pub fn pay_with_remark(
 			origin: OriginFor<T>,
 			recipient: T::AccountId,
 			asset: AssetIdOf<T>,
 			amount: BalanceOf<T>,
+			remark: Vec<u8>,
 		) -> DispatchResultWithPostInfo {
 			let who = ensure_signed(origin)?;
+			// TODO : apply custom logic based on remark
```
stanly-johnson, line 124: Should we have some custom logic to implement here right now or we do that in future? 🤔

olanod: We already have some form of custom user-provided logic with the `FeeHandler`, so one thing we can do is to allow the `apply_fees` method to receive a reference to the `PaymentDetail` that contains the remark

olanod (commented):
Looks good overall 👍🏽 to make it production ready we definitely need to add the restriction on the size of the remark and we might also need to require the user to pay extra depending on the size of the remark, sometimes one way to do it is to reserve funds from the user based on the size of the object they want to store which is returned when the user deletes the data from the chain(e.g. identity pallet), not sure it is what we should do though 🤔 I've been thinking to not keep things in storage forever and do proper clean-up we should be deleting the payment entries once the payment is completed. wdyt?

stanly-johnson:
> Looks good overall 👍🏽 to make it production ready we definitely need to add the restriction on the size of the remark and we might also need to require the user to pay extra depending on the size of the remark, sometimes one way to do it is to reserve funds from the user based on the size of the object they want to store which is returned when the user deletes the data from the chain(e.g. identity pallet), not sure it is what we should do though 🤔 I've been thinking to not keep things in storage forever and do proper clean-up we should be deleting the payment entries once the payment is completed. wdyt?

I think the simplest thing would be to charge a higher incentive to the user using remarks, something like
```rust
let incentive_amount = match remark {
 Some(_) => ((T::IncentivePercentage::get() + Percent::from_percent(2)) * amount),
 None => T::IncentivePercentage::get() * amount,
};
```
but like you said won't do us any benefit unless we delete the storage as soon as we release/cancel ie when the incentive gets refunded. I think deleting the entire payment might be too excessive, since we only store the latest payment for a pair of accounts. Maybe we could just clear the remark 🤔

pallets/payment/src/types.rs
```diff
@@ -77,7 +86,7 @@ pub trait DisputeResolver<Account> {
 }
 
 /// Fee Handler trait that defines how to handle marketplace fees to every payment/swap
-pub trait FeeHandler<Account> {
+pub trait FeeHandler<Account, Remark> {
 	/// Get the distribution of fees to marketplace participants
-	fn apply_fees(from: &Account, to: &Account) -> (Account, Percent);
+	fn apply_fees(from: &Account, to: &Account, remark: &Option<Remark>) -> (Account, Percent);
```
olanod, line 91: It won't be very useful to users of this pallet that implement their own version of the trait to have the remark as a generic type without bounds of any kind so they can use the data somehow, but what bounds should it have? Instead of trying to restrict the remark to follow some specific behavior/trait I'd say we just use the type as is(e.g. `Vec<u8>`) or more useful for developers I think it would be the `PaymentDetail` to have more data to make different decisions. Since we construct the type after we call the `apply_fees` one option could be to make that `fee_detail` property an `Option` to be able to pass the detail to the function. Another option that favors more the encapsulation principle is to have a separate type that is a redacted version of the `PaymentDetail` and has a subset of the fields and then it can be converted to the payment detail.

e.g.
```rust
struct PaymentInfo<Asset, Amount> {
        pub asset: Asset,
	pub amount: Amount,
        pub remark: Vec<u8>,
}

// then some means to convert
impl From<PaymentInfo> for PaymentDetail { ... }
```
... or something like that

olanod (changes requested):
About charging a percentage for the remark I'm not sure, it can get expensive for users trading with costly stuff, then we would indirectly create an incentive for people to not use the feature which is the opposite of what we want since marketplaces would use it. This is the kind of payment that should be minimal part of the "system fee" derived from the weight of the call. For now it's fine, we can revisit this later and see how we can use the weight system to refund the fees when the remark is smaller than the `MaxRemarkLength`

> we only store the latest payment for a pair of accounts

Still can produce a lot of "garbage" in the chain, many people are expected to be trading regularly with different merchants every time so all those payments in cancelled or completed/released state will just clog the chain. Deleting storage entries doesn't mean it won't be available anymore as users can use archive nodes to query the state at a specific point in time, outside consumers(e.g. a valor based merchant API) can keep historical data in a more convenient format to be queried.

pallets/payment/src/types.rs
```diff
@@ -77,7 +86,11 @@ pub trait DisputeResolver<Account> {
...
 
 /// Fee Handler trait that defines how to handle marketplace fees to every payment/swap
-pub trait FeeHandler<Account> {
+pub trait FeeHandler<Asset, Amount, Account> {
 	/// Get the distribution of fees to marketplace participants
-	fn apply_fees(from: &Account, to: &Account) -> (Account, Percent);
+	fn apply_fees(
+		from: &Account,
+		to: &Account,
+		detail: &PaymentDetail<Asset, Amount, Account>,
```
olanod, line 94: mmm this is not gonna scale well if every time we change the payment details we have to change our trait. One last change could be to change payment detail to be generic over a single parameter i.e. `T` that is bound to be `Config` and then we can use `AssetIdOf<T>` and `BalanceOf<T>` inside `PaymentDetail<T>`

stanly-johnson: Not sure if this is what you mean,

```rust
pub struct PaymentDetail<T : Config> {
	/// type of asset used for payment
	pub asset: AssetIdOf<T>,
	/// amount of asset used for payment
	pub amount: BalanceOf<T>,
	/// incentive amount that is credited to creator for resolving
	pub incentive_amount: BalanceOf<T>,
	/// enum to track payment lifecycle [Created, Released, Cancelled, NeedsReview]
	pub state: PaymentState,
	/// account that can settle any disputes created in the payment
	pub resolver_account: T::AccountId,
	/// fee charged and recipient account details
	pub fee_detail: Option<(T::AccountId, BalanceOf<T>)>,
	/// remarks to give context to payment
	pub remark: Option<Vec<u8>>, // TODO : switch to BoundedVec if possible
}
```

But trying this throws the error that TypeInfo cannot be implemented over Config

stanly-johnson: > mmm this is not gonna scale well if every time we change the payment details we have to change our trait. One last change could be to change payment detail to be generic over a single parameter i.e. `T` that is bound to be `Config` and then we can use `AssetIdOf<T>` and `BalanceOf<T>` inside `PaymentDetail<T>`

Not sure if this is what you mean,

```rust
pub struct PaymentDetail<T : Config> {
	/// type of asset used for payment
	pub asset: AssetIdOf<T>,
	/// amount of asset used for payment
	pub amount: BalanceOf<T>,
	/// incentive amount that is credited to creator for resolving
	pub incentive_amount: BalanceOf<T>,
	/// enum to track payment lifecycle [Created, Released, Cancelled, NeedsReview]
	pub state: PaymentState,
	/// account that can settle any disputes created in the payment
	pub resolver_account: T::AccountId,
	/// fee charged and recipient account details
	pub fee_detail: Option<(T::AccountId, BalanceOf<T>)>,
	/// remarks to give context to payment
	pub remark: Option<Vec<u8>>, // TODO : switch to BoundedVec if possible
}
```

But trying this throws the error that TypeInfo cannot be implemented over Config

olanod (approved):

PR #176, opened 2022-01-17
[feat] Payment dispute

### LINKED ISSUE #170: Payment dispute extrinsic
When users cancel a payment we should give sellers some time(e.g. a pallet constant with that is X number of blocks?) to raise a dispute in case they already provided the service or gave the product to the buyer but the buyer tried to get the funds back by cancelling.

stanly-johnson:
resolves #170


pallets/payment/src/lib.rs
```diff
@@ -38,13 +38,21 @@ pub mod pallet {
...
+		type FeeHandler: FeeHandler<
+			AssetIdOf<Self>,
+			BalanceOf<Self>,
+			Self::AccountId,
+			Self::BlockNumber,
+		>;
 		/// Incentive percentage - amount witheld from sender
 		#[pallet::constant]
 		type IncentivePercentage: Get<Percent>;
 		/// Incentive percentage - amount witheld from sender
```
olanod, line 50: maybe good to change this duplicated comment alredy :)

olanod (changes requested):
Looks good to me but missing some tests :)

stanly-johnson:
> Looks good to me but missing some tests :)

@olanod I'm rewriting all tests in next PR (#177), the reason is that benchmarks only work well when using the types we actually use in runtime. So for benchmarks to play well with tests and actual runtime, the tests and runtime have to use the same types, so I'm replacing the use of dummy types with actual types in runtime. Will add the new tests there.

PR #177, opened 2022-01-19
Fix benchmarks and add weights

stanly-johnson:
Changes

- Update benchmarks
- Generate weights file and use correct weights for all extrinsic
- Update tests
- Update documentation
- Added `#[transactional` to all extrinsic
- Address evaluation feedback in https://github.com/w3f/Grant-Milestone-Delivery/pull/344

Requires #176
Next steps #179, #180, #181


pallets/payment/src/weights.rs
```diff
@@ -1,23 +1,25 @@
-//! Autogenerated weights for virto_payment
+
+//! Autogenerated weights for `virto_payment`
```
olanod, line 2: Was this file really autogenerated? what's up with the karura stuff

runtime/Cargo.toml
```diff
@@ -53,6 +53,7 @@ pallet-sudo = { git = "https://github.com/paritytech/substrate", default-feature
 pallet-timestamp = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.13" }
 pallet-transaction-payment = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.13" }
 pallet-transaction-payment-rpc-runtime-api = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.13" }
+pallet-collective = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.13" }
```
olanod, line 56: I thought we didn't need this any more

pallets/payment/src/weights.rs
stanly-johnson: Yes, generated from our benchmarks, the Karura stuff was because I copied the Karura template to create a [benchmark template](fix-benchmarks-weights) but we don't really need it unless we use a CI to autogenerate weights. 

Note : the weight values don't really mean anything right now, they correspond to my local machine, we need to decide on a reference machine and generate weights on that machine. I've created a ticket here https://github.com/virto-network/virto-node/issues/179

runtime/Cargo.toml
stanly-johnson: I don't see any other way to get around [this issue](https://github.com/substrate-developer-hub/substrate-parachain-template/issues/87) that's preventing us from running benchmarks for quite a while.

pallets/payment/src/benchmarking.rs
```diff
@@ -1,14 +1,18 @@
...
+use virto_primitives::Asset;
 
 const SEED: u32 = 0;
-const CURRENCY_ID: u32 = 1u32;
-const INITIAL_AMOUNT: u32 = 100;
-const SOME_AMOUNT: u32 = 80;
+const INITIAL_AMOUNT: u128 = 100;
+const SOME_AMOUNT: u128 = 80;
+
+fn get_currency_id() -> Asset {
```
olanod, line 13: How does this compare to having the `const` as before?

pallets/payment/Cargo.toml
```diff
@@ -21,6 +21,7 @@ sp-std = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v
 frame-benchmarking = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.13", default-features = false, optional = true }
 orml-traits = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.13", default-features = false }
 scale-info = { version = "1.0.0", default-features = false, features = ["derive"] }
+virto-primitives = { default-features = false, path = "../../primitives" }
```
olanod, line 24: I think we can have some duplication in the benchmark file and have a similar enum as our runtime asset definition to avoid bringing this dependency, it would defeat the purpose of https://github.com/virto-network/virto-node/issues/154 to bring this back

README.md
```diff
@@ -85,6 +84,32 @@ Then build the target release with the following command:
...
+```sh
+cargo test --features runtime-benchmarks
+```
+
+### Benchmark
+
+Use the following command to run the benchmark, use the desired pallet name as required:
+
+```sh
+cargo run --release --features=runtime-benchmarks -- benchmark --chain dev  --execution=wasm --wasm-execution compiled --extrinsic="*" --pallet=<pallet_name> --steps=20 --repeat=1 --heap-pages=4096 --output .
```
olanod, line 105: I know our `Makefile` might need some love and updating to be more useful, currently it's a bit complex, but for this is going to be quite useful. Instead of a long command like this one let's just copy it in the Makefile and have a `make bench` or similar that would run this in a more convenient way.

pallets/payment/README.md
```diff
@@ -11,27 +9,67 @@ Each payment gets assigned its own *judge* that can help resolve any disputes be
...
 The RatesProvider module provides implementations for the following traits.
-- [`PaymentHandler`](../../primitives/src/payment.rs)
+- [`PaymentHandler`](./src/types.rs)
+
+## Types 
+
+The `PaymentDetail` struct stores information about the payment/escrow. A "payment" in virto network is similar to an escrow, it is used to guarantee proof of funds and can be released once an agreed upon condition has reached between the payment creator and recipient. The payment lifecycle is tracked using the state field.
+
+```rust 
+pub struct PaymentDetail<Asset, Amount, Account, BlockNumber> {
```
olanod, line 43: This is helpful although it will be annoying to remember to always keep it up to date, it's quite common for this kind of documentation to become outdated fast ... still we can give it a try :)

pallets/payment/Cargo.toml
stanly-johnson: I tried two ways to do it, the duplicate asset definition won't work when we try to generate weights for our runtime since it has to match what we have in runtime. Also tried moving the benchmarks to within the `/runtime` directory. The downside of this is that `cargo test --features runtime-benchmarks` will no longer work. I think we should keep it like this inside our pallet and create a "clean" version of the pallet without benchmarks for orml, or using primitive rust types f(or the purpose of that PR only). I think Acala loosely follows this model since the orml pallets don't have benchmarks, this is left to the teams that add the pallet to the project I believe. Wdyt?

pallets/payment/src/benchmarking.rs
stanly-johnson: Wanted to use `Asset::Default()` as an average case for benchmarks and `const` wouldn't allow it. This was the quickest solution I could think of :D

pallets/payment/Cargo.toml
olanod: Then maybe `optional = true` and add it with the `runtime-benchmarks` feature?

stanly-johnson: I've moved it to `dev-dependencies`, cannot restrict to `runtime-benchmarks` only because we need it for both tests and benchmarks ie we want both `make benchmark` and `cargo test --features runtime-benchmarks` to run

olanod: As long as both work it's fine. Actually better in `dev-dependencies`, I thought it had to go in the normal dependencies for the benchmarks to work

stanly-johnson: Oops my bad, it doesn't work. Adding as optional dependency + runtime-benchmarks will break the tests. Adding in dev-dependencies will break benchmarks :(

olanod (approved):

PR #184, opened 2022-02-02
Use BoundedVec for Remark

stanly-johnson:


pallets/payment/src/lib.rs
```diff
@@ -27,11 +27,13 @@ pub mod pallet {
 	use frame_system::pallet_prelude::*;
 	use orml_traits::{MultiCurrency, MultiReservableCurrency};
 	use sp_runtime::{traits::CheckedAdd, Percent};
+	use sp_std::vec::Vec;
 
 	type BalanceOf<T> =
 		<<T as Config>::Asset as MultiCurrency<<T as frame_system::Config>::AccountId>>::Balance;
 	type AssetIdOf<T> =
 		<<T as Config>::Asset as MultiCurrency<<T as frame_system::Config>::AccountId>>::CurrencyId;
+	type BoundedStringOf<T> = BoundedVec<u8, <T as Config>::MaxRemarkLength>;
```
olanod, line 36: `BoundedDataOf` maybe? it's not always going to be string-like data

olanod (commented):
👍🏽 
I'm just not liking what I mentioned once about the many generic parameters of `PaymentDetail` and related types, every change in the struct is a bit refactor all over the place and when people start using our pallet might become a breaking change every time. We can make it an issue and have it around as reminder

olanod (approved):
Awesome! Glad you found the way around the type info issue, looks much better now

PR #186, opened 2022-02-06
claim_refund extrinsic

### LINKED ISSUE #180: claim_refund extrinsic
Followup to #170 

This extrinsic will allow user to claim the refund if the `recipient` has not created a dispute.

```rust
#[transactional]
#[pallet::weight(T::WeightInfo::claim_refund())]
pub fn claim_refund(
	origin: OriginFor<T>,
	recipient: T::AccountId,
 // check if refund expiry time has passed
 // cancel payment and refund user
}

```

stanly-johnson:
resolves #180


pallets/payment/README.md
```diff
@@ -40,21 +41,21 @@ The RatesProvider module provides implementations for the following traits.
 The `PaymentDetail` struct stores information about the payment/escrow. A "payment" in virto network is similar to an escrow, it is used to guarantee proof of funds and can be released once an agreed upon condition has reached between the payment creator and recipient. The payment lifecycle is tracked using the state field.
 
 ```rust 
-pub struct PaymentDetail<Asset, Amount, Account, BlockNumber> {
+pub struct PaymentDetail<T: pallet::Config> {
```
olanod, line 44: Easy to forget updating the docs isn't it? 😛

olanod (approved):
Code looks good and we can proceed for now but ideally the refund is executed automatically. As usual we can tackle this auto-refund in a follow up issue ;)  
There are several ways to do this but perhaps it will be a good time for off-chain workers to shine? since we might be iterating and filtering the potentially big list of payments I think it's something we can delegate to a worker, we might also want to keep track separately of payments in the `RefundRequested` state for simpler processing 🤔 wdyt?

Issue #174, opened 2022-01-17
Partial refund when resolving disputes


olanod:
Based on a question from @JosephKnecht-lab in his [external review](https://github.com/w3f/Grant-Milestone-Delivery/pull/342#issue-1105724998), we could consider having a method for dispute resolvers to be able to split resources between the buyer and seller in case the current all-or-nothing approach is not enough.

olanod:
I'm thinking to use this as a starting point to do do some refactoring to simplify our API, we can start by first introducing an internal `settle_payment(from: AccountId, recipient: AccountId, recipient_share: Percent)` method that is used by some of our existing extrinsics. E.g. `cancel` is a `settle_payment(from, recipient, Percent::zero())` or `release` a `settle_payment(from, recipient, Percent::one())`. Later instead of `resolve_cancel_payment` and `resolve_release_payment` we expose the `settle_payment` directly.

PR #189, opened 2022-02-10
[refactor] settle_payment function

stanly-johnson:
Removes `release_payment` and `cancel_payment` functions within PaymentHandler trait and replaces it with `settle_payment` function. Only the internal implementation has changed, the extrinsic remain the same.

Part of #174


pallets/payment/src/lib.rs
```diff
@@ -422,95 +445,58 @@ pub mod pallet {
...
+					// unreserve the incentive amount and fees from the owner account
 					match payment.fee_detail {
-						Some((_, fee_amount)) => {
+						Some((fee_recipient, fee_amount)) => {
 							T::Asset::unreserve(
 								payment.asset,
 								&from,
 								payment.incentive_amount + fee_amount,
 							);
+							// if the settlement is a release, transfer fee to marketplace
```
olanod, line 474: I think marketplace should get the payment fee always, or at least when the share is not 0

olanod: also let's not mention marketplace since this pallet doesn't know what that is

olanod (changes requested):
😃

olanod (approved):

PR #190, opened 2022-02-12
Payment request

### LINKED ISSUE #171: Payment request
Haven't thought of this design much but just as users can `pay` others, they should be able to do a `request_payment` that creates a payment in pretty much in the same way, i.e. we create and fill-in a `PaymentDetail` with a `PaymentState::Requested`(or similar) and allow the user receiving the request to confirm. 

One use case is the swap of fiat to crypto, fiat is the off-chain product being sold but the "seller"(someone buying crypto) initiates the trade that the crypto holder can `accept` to proceed with the usual escrow or `accept_and_release`(basically a normal transfer) if there's a trust relation between the two parties(or the initiator is simply to confident).
Another use case this could be good for is subscriptions, someone providing a service can regularly issue the payment request and the user just needs to accept. Perhaps later a separate `subscriptions_pallet` can be the initiator of those requests and have an auto release mechanism if a user decides to trust the merchant.

stanly-johnson:
resolves #171


olanod (changes requested):
Hi! I'm thinking the new extrinsics can share part of the code we already have in the `PaymentHandler` if we split the `create_payment` in two like `create_payment` and `pay`.  
Basically we just take out the part of our existing method that makes the reserves and transfers and put it in `pay`. For the incentive we can add it as an extra parameter of `create_payment` which would be 0 in the `request_payment`, it's still important that the fee detail is there when the user is going to accept the payment request.

pallets/payment/src/lib.rs
```diff
@@ -379,25 +402,84 @@ pub mod pallet {
...
+			charge_incentive_amount: bool,
+		) -> Result<PaymentDetail<T>, sp_runtime::DispatchError> {
 			Payment::<T>::try_mutate(
 				from.clone(),
 				recipient.clone(),
-				|maybe_payment| -> DispatchResult {
+				|maybe_payment| -> Result<PaymentDetail<T>, sp_runtime::DispatchError> {
 					// ensure a payment is not already in process
 					if maybe_payment.is_some() {
 						// do not overwrite an in-process payment!
```
olanod, line 485: Comments like docs are easy to get outdated and out of sync, I believe there's no released/cancelled states anymore? 😜

pallets/payment/src/lib.rs
```diff
@@ -415,43 +497,66 @@ pub mod pallet {
 					}
 					// Calculate incentive amount - this is to insentivise the user to release
 					// the funds once a transaction has been completed
-					let incentive_amount = T::IncentivePercentage::get().mul_floor(amount);
+					let incentive_amount = match charge_incentive_amount {
+						true => T::IncentivePercentage::get().mul_floor(amount),
```
olanod, line 501: I think I'd still prefer to have this incentive be passed explicitly as a parameter and calculate the amount if the percentage passed is not 0. I'm not a big fan of a boolean as function parameters as they make code harder to read and less maintainable since when you see `some_function(true)` you don't really have context and have to dig deeper to see what it actually means.

pallets/payment/src/lib.rs
```diff
@@ -379,25 +402,84 @@ pub mod pallet {
...
-		/// is triggered the amount is released to the recipent and incentive released to creator.
+		/// The function will create a new payment. The fee and incentive amounts will be calculated and the
+		/// `PaymentDetail` will be added to storage.
 		#[require_transactional]
 		fn create_payment(
 			from: T::AccountId,
 			recipient: T::AccountId,
 			asset: AssetIdOf<T>,
 			amount: BalanceOf<T>,
 			remark: Option<BoundedDataOf<T>>,
```
olanod, line 475: just being picky, it's nicer to have optional stuff at the end :)

pallets/payment/src/lib.rs
```diff
@@ -415,43 +497,66 @@ pub mod pallet {
...
 
+		/// The function will reserve the fees+transfer amount from the `from` account. After reserving
+		/// the payment.amount will be transferred to the recipient but will stay in Reserve state.
+		#[require_transactional]
+		fn reserve_payment_amount(
+			from: &T::AccountId,
+			to: &T::AccountId,
+			payment: PaymentDetail<T>,
+		) -> DispatchResult {
+			let fee_amount = match payment.fee_detail {
```
olanod, line 537: wondering if `map` and `unwrap_or_default` would be shorter ... at least they are sexier 😂

pallets/payment/src/lib.rs
```diff
@@ -379,25 +402,84 @@ pub mod pallet {
...
+		/// `PaymentDetail` will be added to storage.
 		#[require_transactional]
 		fn create_payment(
 			from: T::AccountId,
 			recipient: T::AccountId,
 			asset: AssetIdOf<T>,
 			amount: BalanceOf<T>,
 			remark: Option<BoundedDataOf<T>>,
-		) -> DispatchResult {
+			payment_state: PaymentState<T::BlockNumber>,
```
olanod, line 476: we don't have something like a negative balance right? thinking if we can pass a negative amount it means it's a request and we save the extra argument ... but never mind it'd be just to keep my ocd happy 🤣

olanod (commented):
Last picky picks, looks good otherwise :)

olanod (approved):

PR #197, opened 2022-02-23
Combine pay and pay_with_remark

stanly-johnson:
Address feedback from https://github.com/open-web3-stack/open-runtime-module-library/pull/691

- combine pay and pay_with_remark
- add MinPaymentAmount


pallets/payment/src/lib.rs
```diff
@@ -56,6 +56,9 @@ pub mod pallet {
 		/// Buffer period - number of blocks to wait before user can claim canceled payment
 		#[pallet::constant]
 		type CancelBufferBlockLength: Get<Self::BlockNumber>;
+		/// Minimum amount to be transferred in a payment
+		#[pallet::constant]
+		type MinPaymentAmount: Get<BalanceOf<Self>>;
```
olanod, line 61: Don't think it's needed if we use [`MultiCurrency::minimum_balance`](https://github.com/open-web3-stack/open-runtime-module-library/blob/master/traits/src/currency.rs#L42) to determine if the transaction should go through

stanly-johnson: That makes sense, but the downside is that the pallet becomes less generic/configurable. For a use case where the chain requires ExistentialDeposit=0, they will have to fork the pallet. Also I feel its better for visibility to explicitly set the minimum value ie. when the pallet is imported this value is forced to be set and not buried inside as implementation detail. wdyt?

olanod: But then you will need something more like [this](https://github.com/open-web3-stack/open-runtime-module-library/blob/master/tokens/src/lib.rs#L199) where the balance can be determined by token. That said I don't think it's going to be less generic since the minimum balance comes from the already generic `MultiCurrency` trait that the user controls, e.g. we have the `ExistentialDeposits` from the tokens pallet configured in our runtime and if we wanted to we can wrap the tokens pallet in some adapter that overrides the minimum balance requirement.
As I commented in the review, I don't know how the minimum payment helps the spam prevention more than the intrinsic fee of the transaction but for sure we should make sure that a payment made to a new account has the minimum necessary to not be reaped

pallets/payment/src/lib.rs
```diff
@@ -179,12 +162,12 @@ pub mod pallet {
 				amount,
 				PaymentState::Created,
 				T::IncentivePercentage::get(),
-				Some(remark),
+				remark.clone(),
```
olanod, line 165:
```suggestion
				&remark,
```
Following the orml PR discussion, I think it's going to be ok to not bother with the storage of the remark since we already publish it in the created event and only need it while creating the payment to pass it the `FeeHandler`. That means we need to add the extra argument to the `apply_fees` method to accept our `&remark`

olanod: Still don't see the need for `clone()`. Maybe even better than borrowing the option could be to receive an `Option<&[u8]>`(`remark.as_ref().map(Borrow::borrow)`) which is more ergonomic and won't break future user's fee handler code if we decide to change the bounded vec for something else.

olanod (commented):
Looks good although after the removal of the remark from the storage I don't see the big need for the minimum payment. What do you think?

stanly-johnson:
> Looks good although after the removal of the remark from the storage I don't see the big need for the minimum payment. What do you think?

IMO we should keep it in the interest of making the pallet more generic and reusable, and we can set MinPayment to zero in virto_runtime. But if you think its not needed I can remove it.

olanod (approved):

PR #198, opened 2022-02-24
Cancel RefundRequested transactions

stanly-johnson:
resolves #188


pallets/payment/src/lib.rs
```diff
@@ -53,6 +57,9 @@ pub mod pallet {
 		/// Maximum permitted size of `Remark`
 		#[pallet::constant]
 		type MaxRemarkLength: Get<u32>;
+		/// Maximum scheduled tasks per block - this number should be small enough to ensure entire block is not consumed by on_init hook
+		#[pallet::constant]
+		type MaxTasksPerBlock: Get<u32>;
```
olanod, line 62: I think this is not used anymore?

pallets/payment/src/lib.rs
```diff
@@ -126,10 +145,47 @@ pub mod pallet {
...
 
 	#[pallet::hooks]
-	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {}
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		/// Hook that execute when there is leftover space in a block
+		/// This function will look for any pending scheduled tasks that can
+		/// be executed and will process them.
+		fn on_idle(now: T::BlockNumber, mut remaining_weight: Weight) -> Weight {
+			while remaining_weight > T::WeightInfo::cancel() {
+				let task = ScheduledTasks::<T>::iter().next();
```
olanod, line 159: Do you know if every call to next is a new read to the storage? if that's the case we probably need to substract it from the remaining weight. I wonder if there's a better way to go through the list of tasks since we could be potentially spending quite some time every block just skipping tasks 🤔

olanod: Sorry for being indecisive, I'm thinking maybe we should mix `on_initialize` and `on_idle`? in a not so good case imagine merchants have a week or a month to react to the refund and suppose there's a bunch of releases queued but only the few that can actually be executed are at the end of the iteration meaning we can often run out weight before we get to do any refunds.

We could instead have "two queues", first a map that indexes by "future block number" -> [(from, to), ...(MAX)]  where MAX is the bound representing the maximum refunds that can be requested in a block that will likely be a small number, if it gets full we can fail in the `request_refund` with a specific error, it will be up to the front-end to try again and make it nice to the user. This queue is processed in the on_initialize for each block where we keep work to a minimum basically just moving the everything in the bounded vec to the second "ready queue" that is the double map you have here. That way when we iterate here we don't need to check the block number because every task is ready to be processed.

olanod (changes requested):

pallets/payment/src/lib.rs
stanly-johnson: Is the issue the cancel_block timing? Could you take a look at https://github.com/virto-network/virto-node/pull/198/commits/4d2e0cc0a69ced2f9730944ddb68ec714b9a19c7, so I've reduced it to one DB read, then make a Vec sorted by the cancel_block, so if the first one cannot be cancelled we can return all, also helps when we lag behind if the blocks are full. I've benchmarked the weights for reading and writing the `Attributes` storage, will add that to the expected weights if this looks ok

stanly-johnson: Added the cost to read/write the scheduled tasks in https://github.com/virto-network/virto-node/pull/198/commits/75ee30efa435ba86c922615561d315ac8f501f56 wdyt?

PR #209, opened 2022-03-11
Fix auto refund processing same task

olanod:
- The on_idle processing of payment refunds was broken since it always
would always process the same task in the loop.
- Changed to filter out future tasks before sortign since it's
wasteful to sort a potentially large vec with tasks that wont't be
processed anyway.
- Added a mechanism to test busy blocks that can't process tasks


pallets/payment/src/lib.rs
```diff
@@ -154,44 +154,39 @@ pub mod pallet {
...
+					Some((from, to, ScheduledTask { task: Task::Cancel, .. })) => {
+						remaining_weight = remaining_weight.saturating_sub(cancel_weight);
+
+						// process the cancel payment
+						if let Err(_) = <Self as PaymentHandler<T>>::settle_payment(
+							from.clone(),
+							to.clone(),
+							Percent::from_percent(0),
+						) {
+							// panic!("{:?}", e);
```
stanly-johnson, line 182: Maybe we can move it to NeedsReview status here? 🤔

olanod: Not sure there's much to do there, it shouldn't happen but it was happening for example because the payment was already removed so it wouldn't make sense to move to review something that isn't there anymore 🤔

stanly-johnson (approved):

PR #210, opened 2022-03-11
Partial refund for disputes

### LINKED ISSUE #174: Partial refund when resolving disputes

stanly-johnson:
resolves #174 

- combine resolve to single extrinsic
- this allows judge to set variable recipient amounts


olanod (approved):
LGTM 👌🏽

PR #212, opened 2022-03-19
Scheduled tasks : Use BoundedBTreeMap

stanly-johnson:


pallets/payment/src/lib.rs
```diff
@@ -155,45 +155,48 @@ pub mod pallet {
...
-				T::WeightInfo::cancel().saturating_add(T::WeightInfo::remove_task());
-
-			while remaining_weight >= cancel_weight {
-				match task_list.pop() {
-					Some((from, to, ScheduledTask { task: Task::Cancel, .. })) => {
+			// reduce the weight used to read the task list
+			remaining_weight = remaining_weight.saturating_sub(T::WeightInfo::remove_task());
+
+			ScheduledTasks::<T>::mutate(|tasks| {
+				let tmp_tasks = tasks.clone();
```
olanod, line 162: Don't think you need this clone since the `iter` doesn't mutate and `collect` creates a new vec

pallets/payment/src/lib.rs
```diff
@@ -155,45 +155,48 @@ pub mod pallet {
...
+						tasks.remove(&(from.clone(), to.clone()));
 
 						// process the cancel payment
 						if let Err(_) = <Self as PaymentHandler<T>>::settle_payment(
 							from.clone(),
 							to.clone(),
 							Percent::from_percent(0),
 						) {
-							// panic!("{:?}", e);
+							// nothing can be done here
```
olanod, line 189:
```suggestion
							// nothing can be done here. TODO: log a warning
```

pallets/payment/src/lib.rs
```diff
@@ -40,7 +40,14 @@ pub mod pallet {
 	pub type AssetIdOf<T> =
 		<<T as Config>::Asset as MultiCurrency<<T as frame_system::Config>::AccountId>>::CurrencyId;
 	pub type BoundedDataOf<T> = BoundedVec<u8, <T as Config>::MaxRemarkLength>;
+	/// type of ScheduledTask used by the pallet
 	pub type ScheduledTaskOf<T> = ScheduledTask<<T as frame_system::Config>::BlockNumber>;
+	/// list of ScheduledTasks, stored as a BoundedBTreeMap
+	pub type ScheduledTaskList<T> = BoundedBTreeMap<
+		(<T as frame_system::Config>::AccountId, <T as frame_system::Config>::AccountId),
+		ScheduledTaskOf<T>,
+		<T as Config>::MaxRemarkLength,
```
olanod, line 49: `MaxRemarkLength` is a number of bytes and this is a bound for a number of items, might be confusing/misleading to use the same constant that at least would have to be named differently. It's reasonable to expect people configuring them very differently like a remark of 8 bytes but allowing to schedule 1000 tasks or who knows, better to introduce a new one ;)

olanod (changes requested):

pallets/payment/src/lib.rs
stanly-johnson: The issue is this is an immutable borrow but tasks.remove later is a mutable borrow, let me see if it can be handled more elegantly

pallets/payment/src/lib.rs
stanly-johnson: Oops, forgot to change it

pallets/payment/src/lib.rs
olanod: 🤔 the borrow is not needed after constructing the filtered list, maybe `{ scoping it in a block }` helps?

stanly-johnson: I can't think of a way that looks good and makes the borrow checker happy 😅

stanly-johnson: I tried that, the borrow checker does not like it, don't see a way to avoid the clone

olanod (approved):
I guess it's fine to clone since it's a vec of references, maybe can be optimized later but 🤷🏽‍♂️

PR #225, opened 2022-05-23
[feat] Basic Communites pallet

### LINKED ISSUE #224: Basic communities pallet
Let's create a first version of `virto-communities` that  enables registration of new communities that will have a sovereign account.  
Related to https://github.com/virto-network/virto-node/issues/133, a `register` call would at minimum accept a `CommunityId(u32)`(or `CommunityId(u8,u8,u16)`) and a domain name in binary representation(e.g.`BoundedVec<u8, ConstU8<25>>`) that would later be used to verify there is a matrix server associated with the domain.

stanly-johnson:
resolves #224 

Changes : 

- [X] Pallet extrinsics
- [X] Tests
- [x] Benchmarks
- [x] Weights
- [x] Implement fee handler for communities


pallets/communities/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+use scale_info::TypeInfo;
+
+#[derive(Encode, Decode, Clone, PartialEq, Eq, MaxEncodedLen, RuntimeDebug, TypeInfo)]
+#[scale_info(skip_type_params(T))]
+#[codec(mel_bound(T: pallet::Config))]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct Community<T: pallet::Config> {
+	// TODO : Maybe every community can configure an asset?
+	pub controller: T::AccountId,
+	pub population: PopulationOf<T>,
```
olanod, line 25: I think we can keep track of population as a simple counter just for reference, it's not good to have the accounts here as it restricts how many users a business can have or how many people are supposed to live in a place. Instead we can do the inverse and track in the user account data which community/marketplaces they belong to

pallets/communities/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+
+// TODO : Use better representative names
+pub type H3CellIndex = u32;
+pub type H3CellRes = u32;
+pub type H3CellValue = u32;
+
+#[derive(Encode, Decode, Clone, PartialEq, Eq, MaxEncodedLen, RuntimeDebug, TypeInfo)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+// TODO : Use better representative names
+pub struct CommunityId {
```
olanod, line 37: If we keep it as a struct it would be cool to have conversion to/from `u32`

olanod: Also worth having a `validate` method that gets more complex over time

pallets/communities/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+	pub enum Event<T: Config> {
+		/// A new community registered
+		CommunityRegistered(CommunityId),
+	}
+
+	// Errors inform users that something went wrong.
+	#[pallet::error]
+	pub enum Error<T> {
+		/// Cell is already occupied
+		CellAlreadyOccupied,
```
olanod, line 94: Since CommunityIds are not only about cells this can instead be a `InvalidCommunityId` or similar

olanod (changes requested):

pallets/communities/src/impls.rs
```diff
@@ -0,0 +1,26 @@
...
+
+/// FeeHandler trait implementation for Communities
+impl<T: Config> FeeHandler<T> for CommunityFeeHandler {
+	fn apply_fees(
+		_from: &T::AccountId,
+		_to: &T::AccountId,
+		_detail: &PaymentDetail<T>,
+		remark: Option<&[u8]>,
+	) -> FeeRecipientShareList<T> {
+		if let Some(remark) = remark {
```
olanod, line 16: A more useful initial implementation could be to check the validity of the community_id after parsing, if invalid let's charge an arbitrary amount(e.g. 1%) that goes to the pallet id turned into an account(just some random account that is not accessible to users) and if valid and the community is registered then we turn the community_id [into an account](https://paritytech.github.io/substrate/master/sp_runtime/traits/trait.AccountIdConversion.html) that would receive a fee lower than that of an invalid remark(e..g 0.5%)

pallets/communities/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+			// Emit an event.
+			Self::deposit_event(Event::CommunityRegistered(id));
+
+			Ok(().into())
+		}
+
+		/// Join an existing community with given communityid
+		#[pallet::weight(10_000 + T::DbWeight::get().writes(1))]
+		pub fn join(_origin: OriginFor<T>, _id: CommunityId) -> DispatchResultWithPostInfo {
+			todo!()
```
olanod, line 107: Let's start with a simple map of Account => Membership, where membership is perhaps a list of communities the user belongs to.

pallets/communities/src/types.rs
```diff
@@ -0,0 +1,65 @@
...
+}
+
+pub type BaseIndex = u8;
+pub type CategoryIndex = u8;
+pub type InstanceIndex = u16;
+
+/// CommunityId representation
+#[derive(Encode, Decode, Clone, PartialEq, Eq, MaxEncodedLen, RuntimeDebug, Default, TypeInfo)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct CommunityId {
```
olanod, line 26: If we implement [`TypeId`](https://paritytech.github.io/substrate/master/sp_runtime/trait.TypeId.html) for `CommunityId` we get the [`AccountIdConversion`](https://paritytech.github.io/substrate/master/sp_runtime/traits/trait.AccountIdConversion.html#impl-AccountIdConversion%3CT%3E) for free.

pallets/communities/src/types.rs
```diff
@@ -0,0 +1,65 @@
...
+use scale_info::TypeInfo;
+use sp_runtime::traits::Zero;
+use sp_std::{fmt, num::ParseIntError, str::FromStr, vec::Vec};
+
+#[derive(Encode, Decode, Clone, PartialEq, Eq, MaxEncodedLen, RuntimeDebug, TypeInfo)]
+#[scale_info(skip_type_params(T))]
+#[codec(mel_bound(T: Config))]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct Community<T: Config> {
+	// TODO : Maybe every community can configure an asset?
```
olanod, line 13: Communities indeed will have an asset, we can actually provide one automatically using the community id as asset id. Our asset type can have a variant that is a community Id(`VirtoAsset::Community(CommunityId)`). To smooth its usage we would implement `From<CommunityId> for VirtoAsset`

olanod (changes requested):

olanod:
We will revisit the idea later, perhaps using an existing pallet in the ecosystem.

PR #253, opened 2023-06-07
Add pallet proxy to Kreivo Runtime

### LINKED ISSUE #51: Add proxy pallet

darkforest0202:
This PR adds the pallet_proxy to the Kreivo runtime.

Closes #51


runtime/kreivo/src/impls.rs
```diff
@@ -64,3 +66,93 @@ impl PauseXcmExecution for XcmExecutionManager {
...
+	/// selection mechanism.
+	Collator,
+}
+impl Default for ProxyType {
+	fn default() -> Self {
+		Self::Any
+	}
+}
+
+impl InstanceFilter<RuntimeCall> for ProxyType {
```
darkforest0202, line 98: @olanod it's important that you review these filters and you agree with them.

olanod (approved):

olanod:
Filters look good enough, if anything we can review them later when the chain gains more functionality.

Issue #242, opened 2023-06-02
XCM specific configuration.


darkforest0202:
The following XCM configuration is needed for Kreivo and Virto parachains:

- For the native token (DOT/KSM) the relay chain should be recognized as a reserve (not asset hubs).
- For other fungible assets, Asset Hubs parachains should be the reserve.

olanod:
Adding to point 2, should we add the [remote account conversion](https://github.com/paritytech/polkadot/pull/6662) that was added recently so users in our parachain can act on the assets hub with their own accounts instead of every asset being locked in the parachain's sovereign account?

darkforest0202:
Depends on #254

darkforest0202:
Facing an error at the moment when reserve transferring an asset from statemine to kreivo:

 `ERROR: UntrustedReserveLocation`

Investigation is on going.

PR #259, opened 2023-06-15
XCM configuration on Kreivo.

darkforest0202:
This PR adds the following:

- Effectively recognizing Statemine as a valid reserve for assets.
- Paying XCM fees with USDT (id: 1984 and USDT as sufficient)
- Setting XCM Emulator that executes a reserve based transfer:
  1) relay chain native tokens  from the relay chain to Kreivo
  2) USDT from Statemine to  Kreivo paying fees with USDT

Still pending:

- Modify Barrier to Deny reserve transfer to Statemine (`DenyThenTry<DenyNativeTokenReserveTransferToStatemine...`)
- Add a check of balance change of ALICE on kreivo once the native token of relay chain is received.

Related #242


runtime/kreivo/src/xcm_config.rs
```diff
@@ -234,12 +243,12 @@ impl pallet_xcm::Config for Runtime {
 	type SendXcmOrigin = EnsureXcmOrigin<RuntimeOrigin, LocalOriginToLocation>;
 	type XcmRouter = XcmRouter;
 	type ExecuteXcmOrigin = EnsureXcmOrigin<RuntimeOrigin, LocalOriginToLocation>;
-	type XcmExecuteFilter = Nothing;
+	type XcmExecuteFilter = Everything;
```
darkforest0202, line 246: change this to `Nothing` or identify which call should be whitelisted.

runtime/kreivo/src/xcm_config.rs
```diff
@@ -234,12 +243,12 @@ impl pallet_xcm::Config for Runtime {
...
 	// ^ Disable dispatchable execute on the XCM pallet.
 	// Needs to be `Everything` for local testing.
 	type XcmExecutor = XcmExecutor<XcmConfig>;
-	type XcmTeleportFilter = Everything;
-	type XcmReserveTransferFilter = Nothing;
+	type XcmTeleportFilter = Nothing;
+	type XcmReserveTransferFilter = Everything;
```
darkforest0202, line 251: For the future , make a test to assure that asset registry is filtering non desired assets.

xcm-emulator/src/lib.rs
```diff
@@ -0,0 +1,466 @@
+pub use codec::Encode;
```
darkforest0202, line 1: Bring this from cumulus.

darkforest0202: This will be possible in release 0.9.43

runtime/kreivo/src/xcm_config.rs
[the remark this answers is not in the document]
darkforest0202: The asset registry is already taking care of the filter, however we might try implementing some more specific filter. At the moment all parachains are just leaving `Everything` and using an asset registry to allow/disallow assets from other parachains

olanod (approved):

Issue #270, opened 2023-06-23
XCM Trader: Replace `FixedFungibleRate`  by `TakeFirstAssetTrader`


darkforest0202:
There is an altenative of the `FixedFungibleRate` xcm trader that we have now and it's the `TakeFirstAssetTrader`.
This trader has the following characteristics:

- use the first asset of the vector of assets sent in a xcm message and pays the fees with it.
- rely on the weight to fee implementation.
- If the amount does not meet the existencial deposit, it brings the amount to that value at least.

This is the definition: https://github.com/paritytech/cumulus/blob/master/primitives/utility/src/lib.rs#L120C1-L255C2

PR #273, opened 2023-06-29
XCM Trader: Replace FixedFungibleRate by TakeFirstAssetTrader

darkforest0202:
Closes: #270

There is an alternative of the `FixedFungibleRate` xcm trader that we have now and it's the TakeFirstAssetTrader.

This PR is implementing that XCM trader type to match asset hub configuration.


olanod (changes requested):
Looks good, just some minor things :)

darkforest0202:
> Looks good, just some minor things :)

Thanks for applying the fixes :)

PR #274, opened 2023-06-30
Add benchmark overhead and project refactor.

darkforest0202:
_This PR should be reviewed once the PR #273 is merged._

This PR is adding the benchmark overhead to the benchmarking process but it's also contains some considerable changes:

- In order to make to work the bechmark overhead in the node, the virto runtime was sync with kreivo (at this point they are exactly the same).
- As kreivo parachain is pretty much ready for lunching, seedling runtime was removed from the repository. As kreivo is launched with lockdown mode status activated, there are no risk to deploy kreivo as it is (actually we are saving ourselves from one runtime upgrade). Also the limited setup of seedling was making the benchmark overhead harder, for almost no gains.
- pallet `pallet_asset_tx_payment` was added into both runtimes as they were required for the benchmark overhead and also because it will be needed at some point.

@olanod you can just focus on the kreivo runtime related code.


runtime/kreivo/src/lib.rs
```diff
@@ -556,7 +569,8 @@ construct_runtime!(
 		Balances: pallet_balances = 10,
 		TransactionPayment: pallet_transaction_payment = 11,
 		Burner: pallet_burner = 12,
-		Assets: pallet_assets::<KreivoAssetsInstance> = 13,
```
darkforest0202, line 559: For some reason that line complained because unmet trait bounds requirements.

Cargo.toml
```diff
@@ -6,5 +6,8 @@ members = [
     'node',
     'runtime/virto',
     'runtime/kreivo',
-    'virto-xcm-emulator',
+]
+
+exclude = [
+    'virto-xcm-emulator'
```
darkforest0202, line 12: This was bringing issues when building the node with `--features runtime-benchmarks`. For the moment let's leave it like this since it does not change anything from the build perspective.

olanod (approved):

PR #297, opened 2023-09-16
296-enrollingremoving-members-from-a-community

### LINKED ISSUE #296: Enrolling/removing members from a community.
## Types

```rs
pub type SizedField<S> = BoundedVec<u8, S>;
pub type ConstSizedField<const S: u32> = BoundedVec<u8, ConstU32<S>>;
```

## Implementation

- [x] `add_member`: Enroll an account as a community member. In theory, any community member should be able to add a member. However, this can be changed to ensure it is a privileged function.
- [x] `remove_member`: Removes an account as a community member. While enrolling a member into the community can be an action taken by any member, the decision to remove a member should not be taken arbitrarily by any community member. Also, it shouldn't be possible to arbitrarily remove the community admin, as some privileged calls would be impossible to execute thereafter.

## Pull Requests

- #297

pandres95:
Closes #296

### DIFF
--- a/pallets/communities/src/lib.rs
+++ b/pallets/communities/src/lib.rs
@@ -116,8 +116,8 @@
 //!
 //! - `fulfill_challenge`: Submit the challenge proof to validate the
 //!   contribution status of the community.
-//! - `add_member`: Enroll an account as a community member. In theory, any
-//!   community member should be able to add a member. However, this can be
+//! - [`add_member`][10]: Enroll an account as a community member. In theory,
+//!   any community member should be able to add a member. However, this can be
 //!   changed to ensure it is a privileged function.
 //! - `open_proposal`: Creates a proposal to be voted by the community. At
 //! this point, there can only be a single proposal at a time.
@@ -127,7 +127,7 @@
 //!
 //! These functions can be called either by the community _admin_ or
 //! dispatched through an approved proposal. !
-//! - [`set_metadata`][10]: Sets some [`CommunityMetadata`][11] to describe the
+//! - [`set_metadata`][11]: Sets some [`CommunityMetadata`][12] to describe the
 //! community.
 //! - `remove_member`: Removes an account as a community member. While enrolling
 //!   a member into the community can be an action taken by any member, the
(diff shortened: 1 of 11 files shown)


pallets/communities/src/lib.rs
```diff
@@ -386,5 +421,46 @@ pub mod pallet {
...
+		/// would be impossible to execute thereafter.
+		#[pallet::call_index(3)]
+		#[pallet::weight(T::WeightInfo::remove_member())]
+		pub fn remove_member(
+			origin: OriginFor<T>,
+			community_id: T::CommunityId,
+			who: AccountIdLookupOf<T>,
+		) -> DispatchResult {
+			Self::ensure_active(&community_id)?;
+			Self::ensure_privileged(origin, &community_id)?;
```
olanod, line 458: This one should go first, to avoid the DB read if we already know the origin is not signed(or maybe we don't even need to check if the community is active for removing members? 🤔).  

Most times the check of checking the origin is the first thing you do, something I don't love about moving those checks to some inner functions is that is not evident first hand that the check of the origin is being made

pandres95: I see. A possible improvement here is to make more explicit where inner functions imply ensuring some origin. So, here's my proposal:

1. Naturally, moving the `ensure_privileged to the top of the calls` to avoid DB calls (although these origins internally make some DB reads if the origin is a proper `RuntimeOrigin::Signed`).
2. Renaming the functions to be more explicit in mentioning that they make ensuring of origins, like this:

```diff
-			Self::ensure_privileged(origin, &community_id)?;
+			Self::ensure_origin_privileged(origin, &community_id)?;
```

```diff
-			Self::ensure_member(origin, &community_id)?;
+			Self::ensure_origin_member(origin, &community_id)?;
```

Now, when discussing about not checking whether the community is active for removing members, I was thinking about some cases where the community should be frozen (by decision of the same community) or locked (by decision of the network) —maybe those names might be reconsidered as whether defining what's a frozen/locked community— but in those cases there might be a room to disallow the community state to change (maybe a frozen community means there's been an external intervention that requires freezing assets and retaining the state of it, maybe a regulator authority requiring that?).

PR #300, opened 2023-09-19
Handling a community-governed treasury account

pandres95:


pallets/communities/src/functions/registry.rs
```diff
@@ -0,0 +1,48 @@
...
+
+		Ok(())
+	}
+
+	pub(crate) fn do_set_metadata(
+		community_id: &CommunityIdOf<T>,
+		value: types::CommunityMetadata<T>,
+	) -> DispatchResult {
+		<pallet::CommunityMetadata<T>>::try_mutate(community_id, |metadata| {
+			if let Some(metadata) = metadata {
```
olanod, line 36: Both branches of the if-let-else seem to do pretty much the same, setting every field separately here is not much different from just setting the entire struct.

olanod (changes requested):
Looks good, just a small change.
As in the comment bellow, once we have the dispatch mechanism in place we can ditch the methods regarding assets transfers as we could do that from the relevant pallet.

pallets/communities/src/functions/registry.rs
pandres95: Here I tried to make an insert/append kind of behaviour. Which would you think would be a better way to achieve it?

pallets/communities/src/lib.rs
```diff
@@ -462,5 +463,50 @@ pub mod pallet {
 
 			Ok(())
 		}
+
+		/// Transfers an amount of a given asset from the treasury account to a
+		/// beneficiary.
+		#[pallet::call_index(4)]
+		#[pallet::weight(T::WeightInfo::assets_transfer())]
+		pub fn assets_transfer(
```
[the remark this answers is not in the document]
pandres95: Gotcha!

pallets/communities/src/functions/registry.rs
olanod: Just replace the metadata like you have it in the else block `*metadata = Some(value);` assigning all the fields of `value` to the respective fields of `metadata` is semantically the same as replacing entire struct ... with extra steps. It would be worth it if the fields of `value` were optional or needed some validation of assigning some defaults but plainly assigning all the fields has no benefit over just doing what's in line 42

pandres95: Gotcha! Considering the overriding actually occurs on the pallet call, this direct assignment is possible.

PR #301, opened 2023-09-21
Issue community tokens and mark one as sufficient

pandres95:


olanod (commented):
Sorry it took a while, I just couldn't convince myself that exposing methods to manage the assets through this pallet adds any major benefit to justify the extra code. Interacting directly with Assets pallet with the right origins should work for all our use-cases :)

If there's need to do transfers/holds/freezes of assets within this pallet we can still use the fungibles traits and derive the asset id from an established convention. e.g.
```rust
fn community_asset_id(id: &CommunityId, alt_asset: Option<AssetIndexOrNameOrSomeOtherIdentifier>) -> Option<AssetId> {
    let id = alt_asset
        .map(|alt| some_hashing((id, alt)))
        .unwrap_or_else(|| id.into());
    T::Assets::asset_exists(id).then_some(id)
}

olanod:
I'm going to close for the mentioned reasons, if anything we reopen ;)

PR #330, opened 2023-12-24
Polkadot Release v1.5.0 migration

darkforest0202:


cargo.tree
olanod, line 1: Why keep the output of `cargo tree` as a file? a new trend I'm not aware of?

node/src/chain_spec/kreivo.rs
```diff
@@ -23,285 +28,277 @@ fn session_keys(aura: AuraId) -> SessionKeys {
...
-		Some(properties),
-		Extensions {
-			relay_chain: "kusama".into(),
-			para_id: KREIVO_PARA_ID,
-		},
-	)
-}
-
-fn kreivo_live_genesis(
+fn genesis(
```
olanod, line 264: Would this produce the same genesis we used to launch the chain?

node/src/chain_spec/kreivo.rs
```diff
@@ -23,285 +28,277 @@ fn session_keys(aura: AuraId) -> SessionKeys {
...
+		],
+		get_account_id_from_seed::<sr25519::Public>("Alice"),
+		KREIVO_PARA_ID.into(),
+	))
+	.with_protocol_id(DEFAULT_PROTOCOL_ID)
+	.with_properties(properties)
+	.build()
+}
+
+pub fn kreivo_kusama_chain_spec() -> ChainSpec {
```
olanod, line 183: This function should return the "live chainspec" but it's using test data like the `"UNIT"` token and the `test_genesis`. I believe this should keep returning the same chainspec as the one registered in the relay chain.

node/src/chain_spec/kreivo copy.rs
olanod, line 1: Any reason for this "copy"?

node/src/command.rs
```diff
@@ -94,7 +87,7 @@ macro_rules! construct_async_run {
 #[cfg(all(feature = "virto-runtime", not(feature = "kreivo-runtime")))]
```
olanod, line 87: should we remove this including the `virto-runtime` feature?

node/src/command.rs
```diff
@@ -366,18 +347,7 @@ pub fn run() -> Result<()> {
...
-				sp_io::SubstrateHostFunctions,
-				frame_benchmarking::benchmarking::HostFunctions,
-			);
-			construct_async_run!(|components, cli, cmd, _config, runtime| { Ok(cmd.run::<Block, ()>()) })
-		}
-		#[cfg(not(feature = "try-runtime"))]
-		Some(Subcommand::TryRuntime) => Err("Try-runtime was not enabled when building the node. \
-			You can enable it with `--features try-runtime`."
-			.into()),
+		Some(Subcommand::TryRuntime) => Err("The `try-runtime` subcommand has been migrated to a standalone CLI (https://github.com/paritytech/try-runtime-cli). It is no longer being maintained here and will be removed entirely some time after January 2024. Please remove this subcommand from your runtime and use the standalone CLI.".into()),
```
olanod, line 350: Should we follow the suggestion and remove the command entirely?

pallets/communities/Cargo.toml
```diff
@@ -41,6 +41,7 @@ std = [
 	"frame-support/std",
 	"frame-system/std",
 	"pallet-assets/std",
+		"sp-core/std",
```
olanod, line 44:
```suggestion
	"sp-core/std",
```

pallets/payments/Cargo.toml
```diff
@@ -49,6 +49,7 @@ std = [
 	"pallet-preimage/std",
 	"pallet-scheduler/std",
 	"sp-std/std",
+		"sp-core/std",
```
olanod, line 52:
```suggestion
	"sp-core/std",
```

Cargo.toml
```diff
@@ -1,12 +1,10 @@
...
 members = [
 	"node",
-	"runtime/kreivo",
-	"runtime/virto",
-	"common",
+	"runtime/kreivo"
 ]
 
-exclude = ["xcm-emulator"]
+exclude = ["xcm-emulator",	"common"]
```
olanod, line 7: I don't like excluding modules just because they cause trouble with the tests so I could look into this to see why it gives problem.

darkforest0202:
@olanod the PR is still in draft , there are many things that will be fixed :)

node/src/command.rs
darkforest0202: let's make it

pallets/payments/Cargo.toml
darkforest0202: fixed here: https://github.com/virto-network/virto-node/pull/328/files#diff-37e1201c2be9513a444ab4b6e3e6c0111ee2409314f82e2bb06fb59efb027b14L52

olanod (approved):
Approving but we should update in a followup PR the chainspec of the live chain to match the one we used to register the parachain(at least change the `testnet_genesis` usage and `UNIT` token)

PR #336, opened 2024-01-22
Bug payment ids not increasing

hbulgarini:


pallets/payments/src/tests.rs
```diff
@@ -37,25 +29,35 @@ fn build_payment() -> Fees<Test> {
 		Some(remark.as_slice()),
 	);
 
-	assert_eq!(
-		PaymentStore::<Test>::get((SENDER_ACCOUNT, PAYMENT_BENEFICIARY, PAYMENT_ID)).unwrap(),
-		PaymentDetail {
+	if assertion == true {
```
olanod, line 32: It's a bit weird to compare booleans to true/false when they can already be used for the check, instead I would suggest to name the variable in a more descriptive way like `if assert_payment_creation {}`.

pallets/payments/src/tests.rs
```diff
@@ -535,6 +537,26 @@ fn request_payment() {
 	})
 }
 
+#[test]
+fn next_id_works() {
+	new_test_ext().execute_with(|| {
```
olanod, line 542: If the reason for the boolean parameter is to allow this block to fail without considering a failure, I'm wondering if `execute_with_safe` could be used instead 🤔

pallets/payments/src/tests.rs
```diff
@@ -105,7 +107,7 @@ fn check_balance_cancellation() {
 #[test]
 fn test_pay_and_release_works() {
 	new_test_ext().execute_with(|| {
-		let fees: Fees<Test> = build_payment();
+		let fees: Fees<Test> = build_payment(true);
```
olanod, line 110: Passing booleans as-is to function parameters makes it hard to reason about what the parameter does, consider factoring it out to a `const` :)

olanod (approved):

pallets/payments/src/tests.rs
hbulgarini: not exactly, the reason for the boolean is to skip specific balances in the context of one payment test only. So for the payment id test, when i tried the second assertion, the function failed because it was trying to check the balance for the first payment (which was increased by the first execution).

Taking into account this is a test i wouldn't make it over complexify the logic for this helper function.

pallets/payments/src/tests.rs
```diff
@@ -2,14 +2,16 @@ use super::*;
...
-	Payment as PaymentStore,
+	LastId, Payment as PaymentStore,
 };
 use frame_support::{assert_ok, traits::fungibles, weights::constants::WEIGHT_REF_TIME_PER_NANOS};
 use weights::SubstrateWeight;
 
 use sp_runtime::{BoundedVec, Perbill};
 
-fn build_payment() -> Fees<Test> {
+const ASSERT_PAYMENT_ASSERTION: bool = true;
```
olanod, line 12: `ASSERT_PAYMENT_CREATION` perhaps ?

olanod (approved):

pandres95 (commented):
I agree with @olanod with the `ASSET_PAYMENT_ASSERTION` naming. It's confusing, and unless you fully understand the process behind `build_payment` (where this parameter is stated as `assert_payment_creation`), you'd end up missing some context on the tests.

Other than that, it looks good.

pallets/payments/src/tests.rs
hbulgarini: buuuuu, wrong search and replace....

hbulgarini:
> I agree with @olanod with the `ASSET_PAYMENT_ASSERTION` naming. It's confusing, and unless you fully understand the process behind `build_payment` (where this parameter is stated as `assert_payment_creation`), you'd end up missing some context on the tests.
> 
> Other than that, it looks good.

100%, was a wrong search and replace

PR #338, opened 2024-02-13
Use PaymentId from common and simplify extrinsics API

olanod:


pallets/payments/src/lib.rs
```diff
@@ -154,79 +154,61 @@ pub mod pallet {
...
 		_,
-		(
-			NMapKey<Blake2_128Concat, T::AccountId>,
-			NMapKey<Blake2_128Concat, T::AccountId>,
-			NMapKey<Blake2_128Concat, T::PaymentId>,
-		),
+		Blake2_128Concat,
+		// Sender
+		T::AccountId,
+		Twox64Concat,
```
hbulgarini, line 162: why not blake2 as well?

pallets/payments/src/lib.rs
```diff
@@ -154,79 +154,61 @@ pub mod pallet {
...
+	>;
 
 	#[pallet::event]
 	#[pallet::generate_deposit(pub(super) fn deposit_event)]
 	pub enum Event<T: Config> {
 		/// A new payment has been created
 		PaymentCreated {
-			sender: T::AccountId,
-			beneficiary: T::AccountId,
+			payment_id: T::PaymentId,
```
hbulgarini, line 183: just double thinking if its still not useful to display the sender and beneficiaries? i understand the frontends would have these values anyway.

pallets/payments/src/lib.rs
```diff
@@ -253,8 +235,10 @@ pub mod pallet {
 		ReleaseFailed,
 		/// Transfer failed
 		TransferFailed,
-
+		///
```
hbulgarini, line 238:
```suggestion
		/// Storage Value does not exist.
```

pallets/payments/src/lib.rs
```diff
@@ -275,18 +259,17 @@ pub mod pallet {
 		#[pallet::weight(<T as Config>::WeightInfo::pay(remark.as_ref().map(|x| x.len() as u32).unwrap_or(0)))]
 		pub fn pay(
 			origin: OriginFor<T>,
-			beneficiary: AccountIdLookupOf<T>,
+			beneficiary: BeneficiaryOf<T>,
```
hbulgarini, line 262: Is there any reason why we are stepping back from AccountsLookups instead of AccountId? All the frame pallets are relying on AccountsLookups because the conversion capabilities that these type includes.

pallets/payments/src/lib.rs
```diff
@@ -335,94 +311,69 @@ pub mod pallet {
...
+		/// the funds after a configured amount of time that the receiver has to
 		/// react and opose the request
 		#[pallet::call_index(3)]
 		#[pallet::weight(<T as Config>::WeightInfo::request_refund())]
-		pub fn request_refund(
-			origin: OriginFor<T>,
-			beneficiary: AccountIdLookupOf<T>,
-			payment_id: T::PaymentId,
-		) -> DispatchResultWithPostInfo {
+		pub fn request_refund(origin: OriginFor<T>, payment_id: T::PaymentId) -> DispatchResultWithPostInfo {
```
hbulgarini, line 344: is this reformat wanted?

pallets/payments/src/lib.rs
```diff
@@ -433,66 +384,56 @@ pub mod pallet {
...
-			origin: OriginFor<T>,
-			sender: AccountIdLookupOf<T>,
-			payment_id: T::PaymentId,
-		) -> DispatchResultWithPostInfo {
+		pub fn dispute_refund(origin: OriginFor<T>, payment_id: T::PaymentId) -> DispatchResultWithPostInfo {
 			use PaymentState::*;
 			let beneficiary = ensure_signed(origin)?;
-			let sender = T::Lookup::lookup(sender)?;
+			let (sender, b) = PaymentParties::<T>::get(&payment_id)?;
+			ensure!(beneficiary == b, Error::<T>::InvalidPayment);
```
hbulgarini, line 391: Maybe the error would say something more descriptive as invalid beneficiary?

pallets/payments/src/lib.rs
```diff
@@ -433,66 +384,56 @@ pub mod pallet {
...
-					// resolve the dispute
-					let reason = &HoldReason::TransferPayment.into();
-					T::Assets::hold(payment.asset.clone(), reason, &beneficiary, payment.incentive_amount)?;
-
-					payment.state = PaymentState::NeedsReview;
-
-					T::Scheduler::cancel_named(("payment", payment_id).using_encoded(blake2_256))
-				},
-			)?;
+			Payment::<T>::try_mutate(&sender, &payment_id, |maybe_payment| -> Result<_, DispatchError> {
```
hbulgarini, line 393: same with formating.

hbulgarini (commented):
I love the new PaymentId approach, i left some minor comments.

pallets/payments/src/lib.rs
olanod: It was the automatic cargo fmt

pallets/payments/src/lib.rs
olanod: Faster! and there's no way for users to influence the hash of the payment id as it is generated on-chain not from user input

pallets/payments/src/lib.rs
olanod: The idea was to allow for a future where beneficiaries are not only accountIds but also something like a membership NFT. `AccountIdLookupOf<T>` usually refers to a `MultiAddress` and the default lookup process is getting the accountId from `MultiAddress::Id(AccountId)` which is usually the only allowed variant.  
I'll leave the `Beneficiary` thing out of the PR for now as this question makes me realize we could later make use of `MultiAddress::Index` to be used for memberships and give them a first class status across the runtime, we will need a custom `Lookup` config in system and then anywhere where a MultiAddress is used users will be able to select a membership(e.g. a balance transfer to a membership with id 123 can be done easily using `MultiAddress:Index(123)` as destination)

hbulgarini: This sounds reasonable.

olanod:
Made some changes based on review, added the manual encoding/decoding of payment id as u64 and tested the happy path manually(rococo-local zombienet)

PR #359, opened 2024-04-12
Update AssetId to support DOT

olanod:
The new representation allows to represent the Location `.2//polkadot` (parents:2, X1(GlobalConsensus(Polkadot))). We don't need migrations since we only have a `FungibleAssetLocation::Sibling {id, pallet, index}` that encodes the same as the new `FungibleAssetLocation::Sibling(Para)`.


pandres95 (approved):
Storage is preserved, so no migrations are needed.

✅ Tested via Chopsticks.
👍 LGTM.

PR #379, opened 2024-04-29
Handling fee payment for Community Account

johandroid:
First functionality "hardcoded" in the pallet payments, preparing to move the functionality to the pallet "fee-handler"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ zombienet-linux
 # IDEs
 .vscode/
 .idea/
+polkadot-sdk
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,6 +52,7 @@ pallet-burner = { default-features = false, path = "pallets/burner" }
 pallet-payments = { default-features = false, path = "pallets/payments" }
 pallet-communities = { default-features = false, path = "pallets/communities" }
 pallet-communities-manager = { default-features = false, path = "pallets/communities-manager" }
+pallet-fee-handler = { default-features = false, path = "pallets/fee-handler" }
 
 virto-common = { default-features = false, path = "common" }
 runtime-common =  { default-features = false, path = "runtime/common" }
--- a/runtime/kreivo/src/payments.rs
+++ b/runtime/kreivo/src/payments.rs
@@ -4,6 +4,7 @@ use frame_support::traits::EitherOf;
 use frame_system::EnsureSigned;
 use pallet_communities::origin::AsSignedByCommunity;
 use parity_scale_codec::Encode;
+use sp_runtime::traits::AccountIdConversion;
 
 parameter_types! {
 	pub const MaxRemarkLength: u8 = 50;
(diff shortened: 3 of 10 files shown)


runtime/kreivo/src/payments.rs
```diff
@@ -28,29 +29,35 @@ impl pallet_payments::BenchmarkHelper<AccountId, FungibleAssetLocation, Balance>
...
 		_beneficiary: &AccountId,
 		amount: &Balance,
 		_remark: Option<&[u8]>,
 	) -> Fees<Runtime> {
 		let min = <Assets as fungibles::Inspect<AccountId>>::minimum_balance(*asset);
+
+		// Johan toca usar: try_from_sub_account
+		let is_paying_fees:bool = match PalletId::try_from_sub_account::<CommunityId>(sender) {
+			Some((pallet_id, _)) if pallet_id == crate::communities::CommunityPalletId::get() => true,
+			_ => false
```
pandres95, line 49: I guess is the contrary: communities don't pay, while others do.

```suggestion
			Some((pallet_id, _)) if pallet_id == crate::communities::CommunityPalletId::get() => false,
			_ => true
```

.gitignore
```diff
@@ -36,3 +36,4 @@ zombienet-linux
 # IDEs
 .vscode/
 .idea/
+polkadot-sdk
```
pandres95, line 39: Guess it'd be cleaner not to mix repositories in first place. In this case, moving your copy of `polkadot-sdk` out of `virto-node`'s repo seems cleaner than ignoring it.

Cargo.toml
```diff
@@ -52,6 +52,7 @@ pallet-burner = { default-features = false, path = "pallets/burner" }
 pallet-payments = { default-features = false, path = "pallets/payments" }
 pallet-communities = { default-features = false, path = "pallets/communities" }
 pallet-communities-manager = { default-features = false, path = "pallets/communities-manager" }
+pallet-fee-handler = { default-features = false, path = "pallets/fee-handler" }
```
pandres95, line 55: Is this a WIP? It'd be nicer to have it as a separate PR. This way, we can immediately approve the _"hardcoded"_ solution after requested changes, and start reviewing the extension pallet later.

pandres95 (changes requested):
Looks nice overall, clever solution for a specific issue. Also, a good start to understanding how to work out on the new pallet. ~If we move that new pallet to another PR, we can leave it for a further review, and still accept the current specific solution that's outlined in the kreivo config.~

Cargo.toml
pandres95: Nvm. Noticed this is different from #380

runtime/kreivo/src/payments.rs
olanod: This can be caught with proper tests

johandroid:
Wrong behaviour of the proposed code, the branch must be deleted

PR #397, opened 2024-05-10
Update `fc-traits-memberships` dep

pandres95:
See virto-network/frame-contrib#11


olanod:
Let's keep pointing to master but make sure we `cargo update` before a release :)

PR #417, opened 2024-07-22
Kreivo de Paseo

pandres95:


olanod (commented):
Should we remove any mention of Rococo?

pandres95:
We use rococo locally since zombienet with Kusama-local is broken. Let me check if we can fix that part.

pandres95:
Done (finally)!

node/src/chain_spec/kreivo/live.rs
```diff
@@ -0,0 +1,89 @@
...
+#[cfg(not(feature = "paseo"))]
+pub fn chain_spec() -> ChainSpec {
+	ChainSpec::from_json_bytes(include_bytes!("./kreivo_kusama_chainspec.json").as_slice()).unwrap()
+}
+
+#[cfg(feature = "paseo")]
+use sp_core::crypto::Ss58Codec;
+
+#[cfg(feature = "paseo")]
+pub fn chain_spec() -> ChainSpec {
```
olanod, line 12: Since the chain is also live we can do the same and include the json bytes.

PR #469, opened 2025-11-14
feat(kreivo-runtime): Receive valid assets from XCM, and (if these pass the barriers) create an asset class when not existing in the runtime

pandres95:
This pull request handles the creation of asset classes on the `Assets` pallet if these are accepted by the barriers, and don't exist on the pallet registry, so they can be imported into Kreivo without having to handle a proposal for each new added asset.


olanod (approved):
LGTM

PR #481, opened 2026-04-18
feat(kreivo-runtime): update to stable2603

olanod:
## Summary
- Bump all polkadot-sdk dependencies to their `stable2603` versions
- Update frame-contrib and pass-authenticators git dependencies
- Adapt runtime to breaking API changes in XCM, pallet configs, weights, and session APIs

## Breaking changes addressed
- `RuntimeDebug` removed from `sp_runtime`, replaced with std `Debug`
- `pallet_referenda_tracks::Config` reworked with granular origin types (`CreateOrigin`, `GroupManagerCreateOrigin`, `GroupManagerOrigin`, `RemoveGroupOrigin`)
- `MutateTracks` trait removed from frame-contrib, replaced with local trait + `CommunityTracksWrapper`
- XCM `TransactAsset` trait signatures changed (`deposit_asset` now takes `AssetsInHolding` by value)
- `AssetClaims` removed from `xcm_executor::Config`
- `XcmFeesTo32ByteAccount` replaced with `ResolveAssetTo`
- `pallet_revive`: `UnsafeUnstableInterface` removed, `OnBurn` added
- `pallet_pass`: `SpendMatcher`, `CallMatcher`, `MaxFilteredCalls`, `MaxFilteredAssets` added
- `sp-session` v2 API: `generate_session_keys` signature changed
- New weight methods for `pallet_referenda_tracks`, `pallet_assets`, `cumulus_pallet_parachain_system`

## Dependencies updated
- pass-authenticators: virto-network/pass-authenticators@cf85a9b
- frame-contrib: virto-network/frame-contrib@0b576914

## Test plan
- [x] Full workspace compiles (native + wasm)
- [ ] CI checks pass
- [ ] Runtime upgrade tested on zombienet

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #483, opened 2026-09-19
chore(ci): guard against unpinned git dependencies

pandres95:
## Why

Kreivo depends on frame-contrib (and pass-authenticators) through bare `git = "..."` URLs, so the only thing pinning them is `Cargo.lock`. Any `cargo update` can silently move the runtime to a different SDK line or pallet API. That is how `0.17.0-pre.1` and `master` ended up on different frame-contrib states (`stable2512` vs `stable2603` plus the `pallet-pass` per-device call filters), with metadata drift between them.

frame-contrib now has tags for each state Kreivo has shipped:

| Tag | Commit | Used by |
|---|---|---|
| `polkadot-stable2509` | `26630fb` | `0.16.9`, `dev` |
| `polkadot-stable2512` | `306d02b` | `0.17.0-pre.1` |
| `polkadot-stable2603` | `0b57691` | `master` (this PR) |

## What

- **pass-authenticators** is pinned with `rev = "ec1797c…"`, the commit already in `Cargo.lock`. Only the lockfile source string changes; resolution is identical (`cargo metadata --locked` passes).
- **New `pinned-git-deps` lint job:**
  - Every git dependency must have `tag` or `rev`.
  - frame-contrib is the exception, and its pin is enforced on `Cargo.lock`: every `fc-*` crate must resolve to **one** commit, and that commit must be **tagged** in frame-contrib.

### Why frame-contrib isn't pinned with `tag = ...` in `Cargo.toml`

pass-authenticators also depends on `fc-traits-authn` and `fc-pallet-pass` through the bare frame-contrib URL. Cargo treats `…/frame-contrib?tag=X` as a **different source** from the bare URL, so adding `tag` here would compile two copies of those crates and break trait resolution. `[patch]` can't redirect a URL to itself. The lockfile check gives the same guarantee without the duplication.

The long-term fix is to consume frame-contrib from crates.io, where versions unify normally. That work is in virto-network/frame-contrib#77, which adds lockstep versioning and automated crates.io releases.

## Verification

- `cargo metadata --locked` ✅
- Both guard steps run locally: the second one failed before the tags existed and passes now (`0b57691` → `polkadot-stable2603`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pandres95:
Superseded by #484. `master` was reset to `0.17.0-pre.1`, and this branch was built on the previous `master` (#481, now `dev`). #484 carries an updated version of the guard: frame-contrib now comes from crates.io, so the exception for the bare frame-contrib URL is gone, and the lockfile check only applies when frame-contrib is git-sourced (currently just `fc-pallet-black-hole` at `v2.0.0`).

PR #472, opened 2025-12-18
feat: Modifying to fit into asset precompiles

johandroid:
Based on suposition of u64 is enough to decode assets into our Fungible struct, we are modifying it to have at max a u64 decoded asset "address"

PR #485, opened 2026-09-19
feat(kreivo-runtime)!: 0.18 — stable2606-2 and frame-contrib 3.0.0-pre.1 from crates.io

pandres95:
## Context
`dev` now holds what used to be `master` (#481: `stable2603` plus the frame-contrib v3 adoption). This PR is the start of **0.18**. It moves `dev` to polkadot-sdk `stable2606-2` and consumes frame-contrib as crates.io packages instead of through a bare git URL. The 0.17 counterpart is #484.

## Dependencies
- **frame-contrib `3.0.0-pre.1` from crates.io.** This is the next major: the #68, #71 and #72 pallet changes plus `fc-pallet-fees`. See the [changelog](https://github.com/virto-network/frame-contrib/blob/feat/polkadot-sdk-stable2606/CHANGELOG.md).
  - `fc-pallet-black-hole` is pinned to the `v3.0.0-pre.1` tag until its crates.io ownership is sorted out. It depends only on SDK crates, so this doesn't duplicate anything.
- **pass-authenticators** is pinned to `release/frame-contrib-v3` (`c3640de`), which uses the same crates.io frame-contrib.

## stable2606 port
#481 already covered `stable2603`, so only these were needed. Each keeps today's behaviour:
- `sp_core::blake2_256` is no longer re-exported.
- `SchedulingSignatureVerifier = ()`, so V3 scheduling stays off.
- `max_claim_queue_offset = 0`, the collator default on the V1/V2 path.
- `AssetIdAllocator = ()`.

Version `0.18.0`, `spec_version` `135`, plus the `pinned-git-deps` lint job from #484.

## Before this ships
- **`pallet-pass` `SessionKeys` changes value type with no migration** (#71). Existing entries won't decode after the upgrade, so this needs a migration or an explicit storage clear before 0.18 goes out.
- try-runtime against a mainnet snapshot (`check-migration.yml`) should show whether `stable2606` needs more migrations.

## Verification
- `cargo check --locked --workspace --all-targets` ✅
- `kreivo-runtime` with `runtime-benchmarks,try-runtime` ✅
- `cargo +nightly fmt --check` ✅
- `cargo test --locked --workspace`: 49 passed ✅

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pandres95:
## Needs a storage migration before this reaches mainnet

While fixing try-runtime on #484 I found that #472 changed `FungibleAssetLocation`'s variant 2 from `External { network, child }` to the unit `PolkadotNativeDOT`. That shortens the DOT asset id from `02 00 00` to `02`.

Mainnet (0.16.9) stores the old encoding, so a runtime with the new shape can't find those entries: try-runtime reports `"Assets" try_state checks failed: Orphaned Account found`. On Kusama today that is one `Assets::Asset` entry and one `Assets::Account` row; metadata, approvals, holds and freezes have none.

**#484 (0.17) restores the old variant**, so the hotfix upgrades mainnet without touching asset storage. **This line keeps the new shape**, which the asset precompiles need (an external parachain asset doesn't fit in 64 bits with the old shape), so 0.18 needs a migration that rewrites those keys from `02 00 00` to `02` in `Assets::Asset` and `Assets::Account` (plus `Metadata`/`Approvals` if any exist by then), with try-runtime pre/post checks.

Also worth checking before the testnet upgrade: the testnet runs `0.17.0-pre.1`, which already has the new shape, so going from there to a 0.17.0 runtime with the old shape needs the reverse translation (or a reset). Its RPC was down (Cloudflare 522) when I looked, so I couldn't check its state.

pandres95:
## Update: no migration needed

Following up on my comment above. Checking where the `u64` asset-id form is actually required turned up nothing:

- Kreivo doesn't wire up revive precompiles (`type Precompiles = ()`), and revive isn't planned soon.
- `as_u64`/`TryFrom<u64>` had no callers outside tests on either line. `as_u64` also panics for encodings longer than 8 bytes (e.g. an Ethereum-network asset).
- The upstream ERC-20 precompile takes its asset id from the contract address, and `AddressMatcher::Prefix` only leaves **4 bytes** free, not 8. So the packed shape wouldn't have satisfied that interface either; `AssetIdExtractor` is generic, so the fitting design is a 4-byte index resolved through a registry.

So this line now **keeps the on-chain encoding** as well (`External { network, child }`), matching #484, and the `u64` helpers are gone. No chain needs an asset-id migration, and Ethereum-network assets stay representable.

If a compact id is wanted for precompile addresses later, it should be an inner representation translated at that boundary, not the storage key.

